Issue #7, opened 2025-04-16
Allow evaluations while finetuning via torchtune


niznik-dev:
Edit the yaml file for torchtune fine-tuning to allow for periodic evaluation while training and confirm the results are saved somewhere (W&B, log file, etc.)

niznik-dev:
@alessandramaranca This is distinct from #6 right?

msalganik:
I thought that this should be possible.  I asked claude and it suggested some things that look like what we did for last summer
https://claude.ai/share/7457ae7b-7990-45ad-a5d0-5a58e1f34d80

This part looked familiar: 
# Add to your config file
evaluation_strategy: "steps"  # Evaluate at regular intervals
eval_steps: 500  # Evaluate every 500 steps
per_device_eval_batch_size: 64  # Batch size for evaluation

But then I looked in torchtune and I don't see it.  Maybe this is a hallucination.  Or maybe this is how they do it in a different package.  This seems like it should be in torchtune natively.  

What kind of searching have we done to see if this is an existing feature?  If it is not yet maybe there is an issue or pull request?

msalganik:
I'm curious about this so I looked through the issues.  It seems like this is not available out of the box yet.

- https://github.com/pytorch/torchtune/issues/2286
- https://github.com/pytorch/torchtune/pull/2238
- https://github.com/pytorch/torchtune/issues/883
- https://github.com/pytorch/torchtune/issues/1042

But this just got done and seems related
- https://github.com/pytorch/torchtune/pull/2464

I don't know the best way to handle this; I'm not sure I understand the trade-offs

niznik-dev:
Thank you for digging - I did a light search yesterday and did not notice these so I'll start reading to form my opinion.

niznik-dev:
I will attempt this via https://github.com/pytorch/torchtune/pull/2464 and report back

niznik-dev:
I have this working for the use of 2 GPUs and need to start a Pull Request for 1 GPU. I'm attaching the image I emailed previously as I should have just updated this issue 😅 

![Image](https://github.com/user-attachments/assets/d5c5ffbc-5d8d-4b9e-953d-843a2c098a92)

niznik-dev:
I moved this issue to Review because we're not in control of the timeline for when the single GPU validation would be added; I think verifying we can get that working properly with the nightly build of torchtune once the PR is merged in their codebase would be a new issue on our board

niznik-dev:
[...]

PR #11, opened 2025-04-16
Creating a slurm/yaml generator for runs

### LINKED ISSUE #1: Create yaml and slurm generator
Following the example from last summer's work, create a script that can deal with the complexity of changing models, parameters, etc.

This should be a shell script that accepts command line inputs as well as allows simple tweaking of less often changed parameters in the top of the script (not everything needs to be an argument)

niznik-dev:
Closes #1

I used this when working on #7 so I have confidence that we've reached MVP (Minimum Viable Product) and can iterate on the script as needed going forward. Would rather merge it in now before I have too many dependent branches.


niznik-dev:
I've added more changes in #7 but they're directly relevant to getting that working with multiple GPUs so I think this can remain as-is.

alessandramaranca (commented):
Still would like to understand better how we can keep track of runs using this (by name maybe?).

generate_slurm_script.py
```diff
@@ -0,0 +1,84 @@
...
+import yaml
+
+# Skip these when writing the yaml file
+SLURM_ONLY = ['time', 'gpus', 'conda_env', 'account', 'partition', 'constraint']
+
+parser = argparse.ArgumentParser()
+
+# ----- Required YAML Args Reused in Templating -----
+parser.add_argument("--my_wandb_project", type=str, default="PredictingZygosity", help="Project for when results are synced to wandb")
+parser.add_argument("--my_wandb_run_name", type=str, default="MyFirstRun", help="Name for when results are synced to wandb")
```
niznik-dev, line 12: @alessandramaranca Both this and the above will carry over to folder names so it's much easier now. But in terms of the semantics of naming, yes that is an open question. I'm not *overly* concerned like I used to be about adding things like batch size and lora params in the name because you can always get these from looking at either the slurm output file (if you file it with the run) or other files (I think?) in the output directory.

@msalganik Are you still thinking we just generate adjective-noun names?

msalganik: @niznik-dev I personally like the adjective-noun structure.  It makes it easier to talk.  One think I would change is to add the number of adjectives and nouns.  Maybe Claude can help with that?  But since this is going to be open sourced eventually let's avoid adjectives and nouns that might be objectionable to users.

niznik-dev: There are a few packages people have made for doing the above but none are particularly popular from what I can tell. I think asking Claude for some long in-house lists will be fine. I'll add that in a bit.

msalganik (approved):
@niznik-dev once you've added some nouns and adjectives I approve. 

I look forward to more specific feedback after I've had a chance to use it a bit.

niznik-dev:
[45e4714](https://github.com/niznik-dev/predicting-zygosity/pull/11/commits/45e47143b66c9f14402c4fae7d8f0e8946144c7a) adds the run name generator. I told Claude to use neutral or positive sentiment words but did not do a manual check. By default, it will create a random name.

niznik-dev:
@alessandramaranca Are we good on this one now?

niznik-dev:
I know all too well what it's about - I'll resolve before the merge and we can discuss merge conflicts another day.

niznik-dev:
@alessandramaranca It's fixed now - it was a conflict over the conda env name (which the generator now needs it to be <CONDA_ENV> so it can be replaceable by the script). Can you approve when you get a chance?

alessandramaranca (approved):

PR #24, opened 2025-05-15
Update issue and pull request templates

### LINKED ISSUE #23: Add Issue and Pull Request Templates
We seem to be struggling with what exactly we need to document with issues, so I'll go ahead and create a template for issues that we can use and fill in going forward.

niznik-dev:
Closes #23


niznik-dev:
@msalganik @alessandramaranca As I'm working on the pull request templates, if you have input on other things that would be good to remind ourselves to add when creating new issues please comment here. It's not ready for review until I have a PR template as well.

msalganik (approved):
@niznik-dev I don't have strong feelings about this, but I'd like to reserve to right  to revise later after we have used them.

One thing that is not clear to me is the difference between dev and feature.

But I assume that there is one so I'm approving.

niznik-dev:
I will adjust the dev and feature templates so it's clearer (e.g. we could use dev for version upgrade tasks vs feature for the things visible to users)

niznik-dev:
A friend of mine contributes to Athena++ and here's what they have:

Screenshot 2025-05-21 at 1 10 04 PM

I think based on this Bug Report, Feature Request, and *something* (A catch-all for Wiki/Documentation/Maintenance?) would be appropriate. I don't think we'd use the rest. Thoughts? @msalganik?

niznik-dev:
This article shows some good choices that React made for steering people elsewhere for documentation, etc. but I'm not sure we need to do that ourselves: https://rewind.com/blog/best-practices-for-using-github-issues/

It sounds like my "dev" idea needs to adjust to possibly be that catch all but I'm open to other names

niznik-dev:
After doing a bit more reading, I've decided your earlier suggestion of 'refactoring' is actually valid for all the maintenance/documentation tasks in addition to what I usually think of as refactoring, so I think we'll stick with those three (bug, feature request, refactoring). Going to merge this to get it done so I can see what it looks like.

PR #25, opened 2025-05-19
Add Output vs Input/Output Fine-Tuning Test

### LINKED ISSUE #17: clarify training on the input and output vs just training on output
I think this is related to masking.
Do we want the model to try to adjust to predict things in the prompt or just the output?

niznik-dev:
Closes #17

I created a simple test to see what happens when we finetune on output only vs input *and* output. Instructions for how to repeat the test are in the new README.md in the test folder. The input is a 500 line json file that simply capitalizes a word (thanks to Claude for the idea). I ran this for 10 (true-test and false-test) and 25 epochs (true-test-25 and false-test-25) with batch size 1 to see how loss and val_loss differed. You can see the results here: https://wandb.ai/niznik-p-princeton-university/input_training_tests?nw=nwuserniznikp

My prediction was that the loss would be *worse* when also finetuning on the input data but that only appears to be true after some critical point. Additionally, the tokens per second are larger when including the input data but that's simply because it's counting the input data tokens in addition to the output tokens (so when we see low tokens per second, it could simply be that a majority of tokens processed are being excluded so it just appears slow).

I'll comment some questions and thoughts in-line and look forward to discussing this.


generate_slurm_script.py
```diff
@@ -14,15 +14,24 @@
 # ----- Required YAML Args Reused in Templating -----
 parser.add_argument("--my_wandb_project", type=str, default="PredictingZygosity", help="Project for when results are synced to wandb")
 parser.add_argument("--my_wandb_run_name", type=str, help="Name for when results are synced to wandb; if not provided, a random name will be generated")
-parser.add_argument("--input_formatting", type=str, default="raw", help="Name of the folder where your input files are stored within input_dir; useful for multiple formatting styles (e.g. difference vs raw values)")
+parser.add_argument("--input_formatting", type=str, default="raw", help="Name of the folder where your input files are stored within input_dir; useful for multiple formatting styles (e.g. difference vs raw values). If same directory, set to empty string.")
```
niznik-dev, line 17: I'm not sure what to do about this. It might be something we can remove if we just sort input by filename instead of into folders like this ("raw", "diff", etc.). Thoughts?

generate_slurm_script.py
```diff
@@ -14,15 +14,24 @@
...
+parser.add_argument("--dataset_val_filename", type=str, default="val.json", help="Name of the model to use (should be in models_dir)")
 
 parser.add_argument("--output_dir_base", type=str, default="/home/$USER/scratch/", help="Full path to the output file folders (final output folder will be 'zyg_out_' + my_wandb_name within this folder)")
 parser.add_argument("--input_dir_base", type=str, default="/home/$USER/scratch/zyg_in/", help="Full path to the input file folders")
 parser.add_argument("--models_dir", type=str, default="/home/$USER/scratch/torchtune_models/", help="Full path to the model file folders")
 
 # ----- Optional YAML Args -----
+parser.add_argument("--batch_size", type=int, default=4, help="Batch size for training")
+parser.add_argument("--epochs", type=int, default=1, help="Number of epochs to train for")
+parser.add_argument("--save_adapter_weights_only", type=str, default="false", help="Whether to save only the adapter weights (true/false)")
```
niznik-dev, line 29: Could default to true to save space but makes loading the model a bit more complicated. True is great for throw away tests that we don't do any extra analysis on in Python.

generate_slurm_script.py
```diff
@@ -14,15 +14,24 @@
...
 parser.add_argument("--max_steps_per_epoch", type=int, help="Maximum steps per epoch (useful for debugging)")
+parser.add_argument("--log_every_n_steps", type=int, default=5, help="How often to log (in steps)")
 parser.add_argument("--run_val_every_n_steps", type=int, default=50, help="How often to run validation (in steps)")
+parser.add_argument("--dataset_split_point", type=int, help="Percentage of the dataset to use for finetuning")
```
niznik-dev, line 33: The way that torchtune wants us to split our dataset for train/val means we may not want to split our files anymore, but I'm not sure. I'm still a little unclear on whether or not we're doing train/val vs train/val/eval.

templates/finetune_template.yaml
```diff
@@ -9,6 +9,9 @@ my_wandb_project: ""
 my_wandb_run_name: ""
 input_formatting: ""
 
+dataset_filename: ""
```
niznik-dev, line 12: These open up a lot more flexibility going forward for any dataset

tests/input_training/README.md
niznik-dev, line 1: Is what I have here acceptable? We need it to be in two parts anyway. We could also have a script that runs the python script and then sbatch but I worry about errors.

tests/input_training/simpleCapitalization.json
niznik-dev, line 1: This could be something we generate on the fly. I asked Claude for a bunch of words and it kept under-delivering but I cobbled 500 together. If we generate it on the fly, I need to figure out how to pull random (positive to neutral) sentiment words in Python and I don't currently have an API or package for that.

PR #29, opened 2025-05-21
Update Installation and First Run Instructions in README.md

### LINKED ISSUE #28: update first run in readme
The first run from the readme needs to be changed now that #1 is complete.

https://github.com/niznik-dev/predicting-zygosity?tab=readme-ov-file#a-test-run

Also, as we think about building for newcomers, I wonder if we can do something that doesn't require a custom dataset that most folks might not have.  Not sure what would be a good candidate now.

niznik-dev:
Closes #28 

* Prefer one set of installation instructions
* Mention when to use the nightly torchtune build
* Add extra model that we've used
* Split input and run instructions into simple/generic (capitalization dataset, coming soon in #17) vs zygosity (the original)
* Update both of the above to the current state of the codebase

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,10 @@ tune download meta-llama/<model_name> --output-dir <model_dir> --hf-token <hf-to
 
 # Formatting Input Files to JSON
 
+## With capitalization dataset
+
+The file already exists for you (coming soon)!
+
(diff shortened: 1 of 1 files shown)


niznik-dev:
Following our new rules and just merging this in

PR #41, opened 2025-06-06
Add config for save_last_epoch_only

### LINKED ISSUE #39: Flag for disabling model saves until final epoch
# Describe the Feature

We should create a flag that will only save the model output at the final epoch but leave the metric logging in tact

## Related Issues

(None)

# Acceptance Criteria
- [ ] A three epoch run results in only an epoch_2 folder in the output (it's CS numbering starting with 0)
- [ ] A single epoch run still has an epoch_0 folder in the output

niznik-dev:
Closes #39 

# Description

A new config option called save_last_epoch_only has been added to the yaml template (defaults to false). This can be changed via the generator with `--save_last_epoch_only true`.

## New Dependencies

(None)

# Testing Instructions

* Create a run using a test dataset (I'm using predictable_or_not) with `--save_last_epoch_only true` -> only the final epoch (minus 1, computer science counting!) has a save folder with model weights (or adapter weights based on what you chose)
* Create a run using a test dataset (I'm using predictable_or_not) with `--save_last_epoch_only false` -> all epochs have a save folder with model weights (or adapter weights based on what you chose)
  * Repeat the above with the argument left out entirely -> should be the same result (all epochs have a save folder)

### DIFF
--- a/custom_recipes/lora_finetune_single_device_val.py
+++ b/custom_recipes/lora_finetune_single_device_val.py
@@ -861,7 +861,7 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
                         )
                     )
                 else:
-                    log.info(f"Skipping checkpoint save for epoch {curr_epoch + 1} to save disk space.")
+                    log.info(f"Skipping checkpoint save for epoch {curr_epoch + 1}...")
 
     def cleanup(self) -> None:
         self._metric_logger.close()
--- a/custom_recipes/lora_finetune_single_device_val.py
+++ b/custom_recipes/lora_finetune_single_device_val.py
@@ -154,6 +154,7 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         self.global_step = 0
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
         self._save_adapter_weights_only = cfg.get("save_adapter_weights_only", False)
+        self._save_last_epoch_only = cfg.get("save_last_epoch_only", False)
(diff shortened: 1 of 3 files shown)


niznik-dev:
Discovered dc0160a during testing - this was a problem with the new argument and `save_adapter_weights_only`

custom_recipes/lora_finetune_single_device_val.py
```diff
@@ -848,14 +849,19 @@ def train(self) -> None:
...
+                    log.info("Starting checkpoint save...")
+                    self.save_checkpoint(epoch=curr_epoch)
+                    log.info(
+                        "Checkpoint saved in {:.2f} seconds.".format(
+                            time.perf_counter() - start_save_checkpoint
+                        )
                     )
-                )
+                else:
+                    log.info(f"Skipping checkpoint save for epoch {curr_epoch + 1} to save disk space.")
```
niznik-dev, line 864: This message can change, I think this was an auto-suggestion

custom_recipes/lora_finetune_single_device_val.py
```diff
@@ -154,6 +154,7 @@ def __init__(self, cfg: DictConfig) -> None:
         self.global_step = 0
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
         self._save_adapter_weights_only = cfg.get("save_adapter_weights_only", False)
+        self._save_last_epoch_only = cfg.get("save_last_epoch_only", False)
```
niznik-dev, line 157: NB For object oriented programming in Python, starting with a _ means it's a private variable and only used internally in this class

PR #46, opened 2025-06-11
Add fine-tuning custom metrics logic to our custom recipe

### LINKED ISSUE #18: add logging of custom metrics to eval script
For the evaluation script that runs during training, we would like to send custom performance measures to WandB.

For now, let's focus on the case of binary outcome prediction.

~~Here's a list of possible things from scikit learn:
https://scikit-learn.org/stable/api/sklearn.metrics.html~~

For now, let's explore [TorchEval](https://docs.pytorch.org/torcheval/stable/) because it should play well with TorchTune.

Image

Most useful ones to include first are:
- MeanSquaredError https://docs.pytorch.org/torcheval/stable/generated/torcheval.metrics.MeanSquaredError.html#torcheval.metrics.MeanSquaredError
 
Let's pass these to WandB using exactly the function name from TorchEval. 

In an ideal world, this should be done without have to make any changes to torchtune.

niznik-dev:
Closes #18 

# Description

Working with Claude Code (which is amazing by the way), we were able to design a pretty clean way to insert custom metrics into a custom recipe. Considering the options, I'm increasingly convinced this is the cleanest way to accomplish what we want. I'm more concerned about double-checking the logic for the exact metrics that @msalganik asked for, which I should compare against how @alessandramaranca has been calculating things in the eval.py script.

## New Dependencies

None, because Claude convinced me that for what we're doing we actually don't need torch eval...yet

# Testing Instructions

This is TBD - I need to see if these new metrics come out nicely in W&B.

**This will be discussed on Friday at our Unconference**

### DIFF
--- a/custom_recipes/lora_finetune_single_device_val.py
+++ b/custom_recipes/lora_finetune_single_device_val.py
@@ -674,9 +674,20 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
     def _loss_step(self, batch: Dict[str, torch.Tensor]) -> torch.Tensor:
         # Shape [b, s], needed for the loss not the model
         labels = batch.pop("labels")
+        print(batch.keys())
         # run model
         with self.activations_handling_ctx:
             logits = self._model(**batch)
+
+        # Calculate custom metrics before computing loss
+        # We do this here because the logits are needed for custom metrics
+        if self._calculate_custom_metrics:
+            metrics = calculate_custom_metrics(logits, labels, self._tokenizer, self._loss_fn.ignore_index)
+            for metric_name, metric_value in metrics.items():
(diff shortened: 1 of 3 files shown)


msalganik:
@niznik-dev 

Amazing!  I'm happy to talk through to test the logic of the code.  And I agree that we don't need torch eval yet for these particular metrics.  

Do these metrics get passed to W&B?  

Did this actually run successfully? 

This is very exciting.  I look forward to hearing more about it at the unconference.

niznik-dev:
Consider a new issue for Mse: {train, eval, both}

niznik-dev:
Proof that we can upload a custom metric to WandB (just a constant 5 all the time) Screenshot 2025-06-20 at 2 14 30 PM

PR #58, opened 2025-06-23
Make Validation Optional

### LINKED ISSUE #44: Validation should be optional
# Describe the Bug

Currently, we do not allow validation to be disabled. Because this is not yet a feature in the stable torchtune release (and because for some tests runs this may not be necessary), we should allow some way for this to be turned off.

Options:
* If run_val_every_n_steps is missing, remove dataset_val from the yaml file generated
* If run_val_every_n_steps is 0 (or less than 1), remove dataset_val from the yaml file generated
* Do nothing (because dataset_val and run_val_every_n_steps should be ignored if we're not using the nightly build anyway), but there would be no way to disable this once these changes are in the stable release (unless you set a number greater than the number of total steps in a run? Would need to test this)

# Steps to Reproduce

1. Run `python generate_slurm_script.py --my_wandb_project bug_issue --my_wandb_run_name val-bug --input_dir_base /home/niznik/scratch/GitHub/predicting-zygosity/tests/predictable_or_not/   --input_formatting ''   --dataset_filename pp.json   --dataset_val_filename pp.json   --dataset_split_point 90   --conda_env ttenv-nightly   --custom_recipe lora_finetune_single_device_val.py
2. Note that the validation parameters are still present despite a lack of run_val_every_n_steps (which currently has a default value)

# Acceptance Criteria
- [ ] (Will wait on choosing an Option listed above)

niznik-dev:
Closes #44 

# Description

Changed `run_val_every_n_steps` so that it defaults to 0; when 0, all yaml entries related to validation are deleted from the `yaml`; I think this opt-in method is the cleanest solution but I'm open to alternatives

## New Dependencies

(None)

# Testing Instructions

- [ ] Run the generator without a value for `run_val_every_n_steps` and verify no validation-related yaml entries remain
- [ ] Run the generator with a value for `run_val_every_n_steps` and verify the validation-related yaml entries are filled

### DIFF
--- a/generate_slurm_script.py
+++ b/generate_slurm_script.py
@@ -30,7 +30,7 @@ parser.add_argument("--save_adapter_weights_only", type=str, default="false", he
 parser.add_argument("--save_last_epoch_only", type=str, default="false", help="Whether to save only the last epoch (true/false)")
 parser.add_argument("--max_steps_per_epoch", type=int, help="Maximum steps per epoch (useful for debugging)")
 parser.add_argument("--log_every_n_steps", type=int, default=5, help="How often to log (in steps)")
-parser.add_argument("--run_val_every_n_steps", type=int, default=50, help="How often to run validation (in steps)")
+parser.add_argument("--run_val_every_n_steps", type=int, default=0, help="How often to run validation (in steps)")
 parser.add_argument("--dataset_split_point", type=int, default=80, help="Percentage of the dataset to use for finetuning")
 parser.add_argument("--train_on_input", type=str, default="false", help="Whether to train on the input data (true/false)")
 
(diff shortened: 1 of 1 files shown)


niznik-dev:
@alessandramaranca @msalganik do either of you have strong opinions about this? I would like to merge this one quickly because it's an ongoing pain point for my tests

msalganik:
Interesting.  Let me see if I understand.

It seems like there are two questions:
- do you want to run validations within each epoch?
- If yes, how often and how is that specified by the user?

I remember this came up a lot this past summer.  I remember thinking that a lot of the confusion came from try to force two choices into option.  

Here's how we did it.  Not saying it has to be the same.
Two options:
--run_validation True 
--additional_validations_per_epoch 2 

The idea is that as a researcher it is easier to specify how many validations you want per epoch rather than how many steps per evaluation.  This requires a bit of algebra that we worked out and I can share with you.

One option that I'm fine with is doing what you have now and then revisiting the exact way to specify the options once we've had some more experience with the code.  What you've proposed seems totally reasonable.

niznik-dev:
#59

PR #60, opened 2025-06-25
Add p(0), p(1), and their Sum as Metrics

### LINKED ISSUE #57: Add p(0) + p(1) = X as a metric
# Describe the Feature

This ticket is specifically to implement the p(0) + p(1) custom metric. I learned when working on this that:

- The list of tensors that is returned from the model is 8 subdivisions of the full sequence length (in this case, 29 each for a total of 29 * 8 = 232 tokens)
- Because our input is always 216 tokens long (for the fake twin data), we can hardcode that we want the logits at position 216 (but this is a terrible hardcode so we need a better way of attacking this longterm)

## Related Issues

#18 

# Acceptance Criteria
- [ ] During any test run of finetuning with an existent `finetune_custom_metrics.py`, this metric (name to be finalized) appears on the WandB output
- [ ] We develop a way to determine the sequence length without hardcoding (i.e. there is code that can determine 216 for this test case and is reliable if the input sequence varies - test with capitalization dataset?)

niznik-dev:
Closes #57 

# Description

Figured out a way to extract our prediction probabilities and add them to custom metrics.

Screenshot 2025-06-25 at 2 16 55 PM

Right now p(0) and p(1) will be more or less noisy based on batch size, because we're getting the average across a batch. The image I posted was for a batch size of 4 - this is less noisy when I use a batch size of 16 for example.

## New Dependencies

(None)

# Testing Instructions

Perform a run with the fake zygosity data and verify that p(0), p(1), and their sum are logged to WandB.

### DIFF
--- a/utils/finetune_custom_metrics.py
+++ b/utils/finetune_custom_metrics.py
@@ -59,16 +59,13 @@ def calculate_prob_0_1_sum(
     # Concatenate along sequence dimension and take logits from where we think the answer is generated
     full_logits = torch.cat(logits, dim=1)
 
-    # Target positions are the last non-ignore index in each sequence
-    # Based on investigation, our prediciton will come 3 later
+    # Target positions are the first non-ignore index in each sequence plus 3
+    # Based on investigation, our prediciton will come 3 later than the first non-ignore index
     # (0: BOS, 1: "Assistant", 2: EOS, 3: "/n" -> "0" or "1")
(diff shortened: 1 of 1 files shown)


niznik-dev:
Working on this has opened up some new architecture questions for me but I think we can save that for a future issue once we start getting this in the pipeline

Issue #53, opened 2025-06-19
Evaluation of subgroup predictability heterogeneity with Inspect AI


alessandramaranca:
# Feature Request: Evaluation of subgroup predictability heterogeneity with Inspect AI

Add an evaluation script to the pipeline that leverages Inspect AI (https://github.com/UKGovernmentBEIS/inspect_ai) to compute group-level evaluation metrics for a given model (base or finetuned). Prediction heterogeneity seems to be a big part of understanding what the model is doing (in particular when the groups that are being predicted better differ from other prediction methods).

# Desiderata 
This is one way of doing this, I'm still not sure the best way is (or how we could potentially directly integrate with hugging face data), 

* Take as input a file with two columns:
  * `"input"`: the prompt string to feed into the model.
  * `"group"`: the group identifier (example: "first bit is 0", or sex, age. Maybe start with the binary case but if it's easy to generalize this we might as well).
* Use Inspect AI to evaluate the model's probabilistic predictions for the next token (i.e., the probability that the next token is `"0"` or `"1"`).
* Compute:
  * Per-sample probabilities for `"0"` and `"1"`.
  * Group-level evaluation metrics (accuracy, AUC, etc.) for each unique value in the `"group"` column.
* Output results summarizing group-level metrics and optionally individual prediction records.

# Tasks

* [ ] Write a new evaluation script (e.g., `eval_with_inspect.py`) that wraps Inspect AI for group analysis.
* [ ] Add CLI args for: input, base model, adapter, and output directory. 
* [ ] Integrate the tokenizer and model logic to extract P(0) and P(1). 
* [ ] Compute metrics grouped by `"group"` column.
* [ ] Write results

msalganik:
@alessandramaranca this looks great.  One thing I would add is that we might want to extend this to outcomes beyond just P(0) and P(1).  That is what we are doing now, but what if we want to just continue the sequence or extract the reasoning that is used by the model.  I'm not saying you need to make it this general now, but it would be good to leave it flexible in the output.

PR #84, opened 2025-08-20
Heterogeneity analysis feature for pre-computed predictions

DonggyuBan:
Relates to #53

# Description

This pull request introduces a heterogeneity detection feature to identify bias in a model's binary prediction outcomes across demographic groups. The feature analyzes a CSV file of pre-computed predictions to detect heterogeneity by group using accuracy and AUC.

```heterogeneity_report.py```: loads a CSV file with a configurable group column, calculates per-group accuracy and AUC, and uses statistical tests to find heterogeneity.

```heterogeneity_eval.py```: converts the CSV file into the Inspect AI format and returns standardized metrics from the analysis.

## New Dependencies

(None)

# Testing Instructions

Run ```heterogeneity_eval.py``` with pre-computed CSV file. 
Use: 
```inspect eval heterogeneity_eval.py -- --input_file file_name.csv --group_column GROUP_NAME```

### DIFF
--- a/heterogeneity/heterogeneity_eval.py
+++ b/heterogeneity/heterogeneity_eval.py
@@ -1,5 +1,9 @@
 import os
+import sys
+import argparse
+
 import pandas as pd
+
 from inspect_ai import Task, task
 from inspect_ai.dataset import Sample
 from inspect_ai.scorer import Scorer, scorer, Score, metric
@@ -22,6 +26,23 @@ class _MetricState:
         self.expected_samples = None
 
 _state = _MetricState()
+
+if '--' not in sys.argv:
+    raise ValueError("Arguments must be passed after '--'")
+
+seperator_index = sys.argv.index('--')
+args_to_parse = sys.argv[seperator_index + 1:]
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--input_file', type=str, required=True, 
+                   help='Path to the input CSV file')
+parser.add_argument('--group_column', type=str, required=True,
+                   help='Column name for group identifiers')
+
+args, unknown = parser.parse_known_args(args_to_parse)
+
+INPUT_FILE = args.input_file
+GROUP_COLUMN = args.group_column
 
 def load_samples_for_inspect(csv_path, group_column='GROUP'):
     """Convert CSV to Inspect AI Samples - works with ANY group column"""
@@ -145,16 +166,13 @@ def heterogeneity_scorer():
 @task
 def heterogeneity_task():
     _state.reset()
-
-    input_file = os.environ.get('INPUT_FILE', 'predictions.csv')
-    group_column = os.environ.get('GROUP_COLUMN', 'GROUP')
     
     print(f"Running heterogeneity task with:")
-    print(f"  Input file: {input_file}")
-    print(f"  Group column: {group_column}")
+    print(f"  Input file: {INPUT_FILE}")
+    print(f"  Group column: {GROUP_COLUMN}")
     
     return Task(
-        dataset=load_samples_for_inspect(input_file, group_column),
+        dataset=load_samples_for_inspect(INPUT_FILE, GROUP_COLUMN),
         solver=[],
         scorer=heterogeneity_scorer()
     )
--- a/heterogeneity/heterogeneity_report.py
+++ b/heterogeneity/heterogeneity_report.py
@@ -246,7 +246,7 @@ def run_analysis(input_file, group_column='GROUP', output_dir='results'):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Analyze model predictions for heterogeneity")
     parser.add_argument('--input_file', type=str, required=True)
-    parser.add_argument('--group_column', type=str, default='GROUP')
+    parser.add_argument('--group_column', type=str, required=True)
     parser.add_argument('--output_dir', type=str, default='results')
     
     args = parser.parse_args()

### CODE REVIEW

File: heterogeneity/heterogeneity_eval.py [hunk 1]
```diff
@@ -0,0 +1,158 @@
...
+
+from heterogeneity_report import (
+    calculate_group_metrics,
+    find_heterogeneity,
+    identify_groups,
+    visualizations,
+    generate_report
+)
+
+accumulated_data = []
```
niznik-dev, line 15: This is not ideal - I think we can reconsider how we handle this list.

For example, can we define it in a task and then pass it to the relevant functions instead? I tend to reserve globals for constants.

heterogeneity/heterogeneity_eval.py
```diff
@@ -0,0 +1,158 @@
...
+EXPECTED_SAMPLES = None
```
niznik-dev, line 16: Similarly, using all caps for this variable name implies it is a constant, but I see it gets changed below. This too is not ideal.

niznik-dev: But I'll let you respond because it may be we're locked into this based on how inspect_ai works.

heterogeneity/heterogeneity_eval.py
```diff
@@ -0,0 +1,158 @@
...
+            return metrics_dict
+        
+        else:
+            return {'samples_processed': float(total_samples)}
+    
+    return compute
+
+@scorer(metrics=[heterogeneity_metric()])
+def heterogeneity_scorer():
+
```
niznik-dev, line 124: I would remove this empty line

heterogeneity/heterogeneity_eval.py
```diff
@@ -0,0 +1,158 @@
...
+        else:
+            return {'samples_processed': float(total_samples)}
+    
+    return compute
+
+@scorer(metrics=[heterogeneity_metric()])
+def heterogeneity_scorer():
+    
+    async def score(state, target, **kwargs):
+        if hasattr(target, 'target'):
```
niznik-dev, line 126: That there are three possibilities here is quite confusing - can you explain what's going on? We may be able to simplify this.

heterogeneity/heterogeneity_eval.py
```diff
@@ -0,0 +1,158 @@
...
+    print(f"Loading data from {csv_path}...")
+    df = pd.read_csv(csv_path)
+    
+    EXPECTED_SAMPLES = len(df)
+    
+    print(f"Loaded {len(df)} rows.")
+    print(f"Using group column: {group_column}")
+    
+    unique_groups = df[group_column].unique()
+    n_groups = len(unique_groups)
```
niznik-dev, line 31: [minor] I would probably call this `n_unique_groups`

heterogeneity/heterogeneity_eval.py
```diff
@@ -0,0 +1,158 @@
...
+
+@metric
+def heterogeneity_metric():
+    def compute(scores):
+        global accumulated_data, EXPECTED_SAMPLES
+        
+        new_scores = scores[len(accumulated_data):]
+        
+        for score in new_scores:
+            if score.metadata:
```
niznik-dev, line 66: When would a score not have metadata?

heterogeneity/heterogeneity_eval.py
```diff
@@ -0,0 +1,158 @@
...
+                'group': state.metadata.get("group")
+            }
+        )
+    
+    return score
+
+@task
+def heterogeneity_task():
+
+    input_file = os.environ.get('INPUT_FILE', 'predictions.csv')
```
niznik-dev, line 147: I'm a bit skeptical that these should be env variables but we can discuss

heterogeneity/heterogeneity_report.py
```diff
@@ -0,0 +1,248 @@
+import pandas as pd
```
niznik-dev, line 1: Same comment on import organization

heterogeneity/heterogeneity_report.py
```diff
@@ -0,0 +1,248 @@
...
+import argparse
+import os
+from sklearn.metrics import accuracy_score, roc_auc_score 
+from scipy.stats import f_oneway
+
+def load_data(file_path, group_column='GROUP'):
+    """Load CSV data for analysis"""
+    df = pd.read_csv(file_path)
+    
+    if group_column != 'GROUP' and group_column in df.columns:
```
niznik-dev, line 14: What happens if it's not 'GROUP' and `group_column` isn't in `df.columns`?

heterogeneity/heterogeneity_report.py
```diff
@@ -0,0 +1,248 @@
...
+        df['GROUP'] = df[group_column]
```
niznik-dev, line 15: [minor] This isn't the most memory efficient way to accomplish this, but I think it's fine for our purposes

heterogeneity/heterogeneity_report.py
```diff
@@ -0,0 +1,248 @@
...
+    
+    for group in df['GROUP'].unique():
+        group_df = df[df['GROUP'] == group]
+        
+        accuracy = accuracy_score(
+            group_df['TRUE_LABEL'],
+            group_df['PREDICTION']
+        )
+        
+        try:
```
niznik-dev, line 32: Oh I'm so glad to see try/except and not assert!

heterogeneity/heterogeneity_report.py
```diff
@@ -0,0 +1,248 @@
...
+                    'z_score': 0
+                })
+                auc_outlier = True
+        
+        if acc_outlier and auc_outlier:
+            identified_groups['outlying_in_both'].append(group)
+    
+    return identified_groups
+
+def visualizations(group_metrics, output_dir='results'):
```
niznik-dev, line 147: I knew there was something I didn't like about this when I saw it called in `run_analysis()`, and now I know what it is. This should be a verb so it reads better. Maybe `visualize_data`?

niznik-dev: Or `create_performance_visualizations` as is below could work - up to you

heterogeneity/heterogeneity_report.py
```diff
@@ -0,0 +1,248 @@
...
+            'heterogeneity_found': bool(heterogeneity_results['accuracy_heterogeneity']['heterogeneity'] or 
+                                 heterogeneity_results['auc_heterogeneity']['heterogeneity']),
+        },
+        'heterogeneity_analysis': heterogeneity_results,
+        'identified_groups': identified_groups,
+        'group_metrics': group_metrics_json
+    }
+    
+    with open(f'{output_dir}/heterogeneity_report.json', 'w') as f:
+        json.dump(report, f, indent=2, default=str)
```
niznik-dev, line 219: Can you double-check the order this prints the values? I know I've needed to disable auto-alphabetizing behavior before when doing `json.dump` to make things a bit more readable.

heterogeneity/heterogeneity_report.py
```diff
@@ -0,0 +1,248 @@
...
+    
+    group_metrics = calculate_group_metrics(df)
+    heterogeneity_results = find_heterogeneity(df, group_metrics)
+    identified_groups = identify_groups(group_metrics)
+    
+    visualizations(group_metrics, output_dir)
+    report = generate_report(df, group_metrics, heterogeneity_results, identified_groups, output_dir)
+    
+    print(f"\nAnalysis complete. Results saved to {output_dir}/")
+    return report
```
niznik-dev, line 239: This is superfluous, right? You shouldn't need to return anything since it's unused.

niznik-dev (changes requested):
I think this is heading in the right direction. Some topics to discuss though before I'd approve it. Well done in general (from what I can follow!).

heterogeneity/heterogeneity_eval.py
DonggyuBan: Seems like the constraint with inspect_ai is that its metrics only accept a scores parameter. I tried using a class _MetricState to organize things better. It still uses global state, but seems cleaner than just global variables.

heterogeneity/heterogeneity_eval.py
DonggyuBan: Fixed it

heterogeneity/heterogeneity_report.py
DonggyuBan: Added error handling code!

heterogeneity/heterogeneity_eval.py
DonggyuBan: I tried using argparse, but Inspect AI doesn't seem to allow it. Should I edit the script to use a configuration file instead?

heterogeneity/heterogeneity_eval.py
```diff
@@ -1,40 +1,47 @@
...
     find_heterogeneity,
     identify_groups,
-    visualizations,
+    visualize_data,
     generate_report
 )
 
-accumulated_data = []
-EXPECTED_SAMPLES = None
+class _MetricState:
```
niznik-dev, line 15: I think I can get behind this, good idea

niznik-dev:
I think we're just down to very minor import optimization and a config file vs environment variables. Well, and probably @alessandramaranca approving as well.

PR #330, opened 2026-02-23
Add numeric risk scoring and fix inverted probabilities

### LINKED ISSUE #311: Support verbalized risk scores for evaluation metrics
### Summary

`folktexts` uses two different metrics of risk: the multiple choice case (calculated directly from next-token probabilities for the positive class) and the numeric case (the prompt directly elicits confidence from model by asking for output as "probability of X class") 

The `risk_scorer.py` in `cruijff_kit` computes risk scores for the first case. We should create a numeric scorer in `tools/inspect/scorers/` that works for the second case - the downstream evaluation (`calibration_metrics.py`) should not have to be changed since both scorers should output the same schema. 

### Suggested implementation 

(suggested by me and Claude)

Create a new scorer `numeric_risk_scorer` in `tools/inspect/scorers/` that:

1. Parses the model's completion text into a float in [0, 1]:
Parses the model's completion text into a float in [0, 1]:
- 1-2 forward passes (tokens), decoded as `0.{digit1}{digit2}` → float
2. Outputs the same metadata schema as risk_scorer:  
```
Score(
      value=CORRECT or INCORRECT,
      answer=...,
      metadata={
          "risk_score": float,  # parsed probability in [0, 1]
          "target": str,        # ground truth label
      }
  )
```
3. Registers the same downstream metrics (expected_calibration_error, brier_score, auc_score) so that evaluation logs are directly comparable between logprob-based and numeric elicitation experiments.

sarahepedersen:
Closes #{311}

# Description

- Adds `numeric_risk_scorer` that will compute all downstream metrics (Accuracy, ECE, etc.) when the model is prompting for verbalized risk ("What is the probability of _____") instead of using the next-token logprobs 
- The risk scorer still requires the ground-truth positive and negative classes in the evaluation setup
- Fixes bug in which we were computing risk scores w.r.t. probability of 0 class instead of 1 (should always be with respect to positive class, so now we proxy that as the last option probability token) 

## New Dependencies

NA

# Testing Instructions

see `/scratch/gpfs/MSALGANIK/sarahep/ck-experiments/acs_income_numeric_risk_2026-02-23/Llama-3.1-8B-Instruct_base/eval/logs` in inspect for performance using numeric risk scores

### DIFF
--- a/.claude/skills/design-experiment/references/scorers.md
+++ b/.claude/skills/design-experiment/references/scorers.md
@@ -77,6 +77,44 @@ Extracts risk scores from logprobs of the first generated token. Computes normal
 
 Note: For multiclass tasks, `risk_score` will be `null` — use `option_probs` instead.
 
+### numeric_risk_scorer
+
+Parses a numeric probability from the model's text output (e.g., "0.73"). For tasks where the model directly outputs a risk score as text rather than via logprobs.
+
+**Source:** `tools/inspect/scorers/numeric_risk_scorer.py`
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `labels` | `("0", "1")` | The ground-truth target values in the dataset: (negative, positive). Used to determine correctness and compute calibration metrics. |
+
+**How it works:**
+1. Parses the model's text output as a float in [0, 1] — this is the `risk_score` (probability of the positive class)
+2. Builds synthetic `option_probs`: `{negative_label: 1-risk, positive_label: risk}`
+3. Thresholds at 0.5 to determine correctness: if risk >= 0.5, the predicted label is `labels[-1]` (positive); this is compared to `target.text`
+
+**Metrics produced:**
+- `risk_score`: The parsed probability value
+- `option_probs`: Synthetic distribution over the two ground-truth labels
+- `CORRECT`/`INCORRECT` based on thresholding at 0.5
+
+**Requirements:**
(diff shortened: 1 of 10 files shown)


.claude/skills/design-experiment/references/scorers.md
```diff
@@ -77,13 +77,52 @@ Extracts risk scores from logprobs of the first generated token. Computes normal
...
+**Source:** `tools/inspect/scorers/numeric_risk_scorer.py`
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `labels` | `("0", "1")` | The ground-truth target values in the dataset: (negative, positive). Used to determine correctness and compute calibration metrics. |
+
+**How it works:**
+1. Parses the model's text output as a float in [0, 1] — this is the `risk_score` (probability of the positive class)
+2. Builds synthetic `option_probs`: `{negative_label: 1-risk, positive_label: risk}`
+3. Thresholds at 0.5 to determine correctness: if risk >= 0.5, the predicted label is `labels[-1]` (positive); this is compared to `target.text`
```
niznik-dev: This is probably nonsense, but thinking back to math class is there an issue with 0 - 1 inclusive and 0.5 being our cutoff? As opposed to, say, 0 - 0.4999... and 0.5 - 0.9999...?

niznik-dev: "math class" 😆

Issue #297, opened 2026-02-09
Remove baseline separation from analyze-experiment reports


niznik-dev:
Undo the separate baseline treatment in analyze-experiment. Base models should appear as regular rows alongside fine-tuned models. The current separation is too vague to determine in many cases.

Discussed in 2026-02-09 software meeting.

PR #333, opened 2026-02-24
Remove base_recipe and stale baseline references from skill docs

### LINKED ISSUE #332: base_recipe silently overrides gradient_accumulation_steps (and other defaults)
## Summary

`base_recipe` (added in PR #260) silently overrides our template defaults with torchtune's upstream recipe values. This caused `gradient_accumulation_steps` to be `8` instead of the expected `1`, leading to an effective batch size 8x larger than intended in recent experiments.

## What happened

In the `acs_income_bs32_2026-02-23` experiment, we set `batch_size: 32` to test GPU memory utilization. But the effective batch size was actually **256** (32 × 8 gradient accumulation steps), giving only 31 steps per epoch. Combined with `num_warmup_steps: 100`, the model never exited warmup — epoch 0 had 0% accuracy, epoch 1 had 64.1% (vs 74.1% at bs=8).

## Root cause

Our `finetune_template.yaml` correctly sets `gradient_accumulation_steps: 1`. But when `base_recipe: "llama3_2/1B_lora_single_device"` is specified in experiment_summary.yaml, `setup_finetune.py` loads torchtune's upstream recipe YAML via `tune cp`, which has `gradient_accumulation_steps: 8`. The 4-level precedence system (CLI > config_file > recipe > argparse_default) lets the recipe value win because we don't explicitly set `gradient_accumulation_steps` in our experiment controls.

**Precedence chain:**
1. CLI args — not set
2. `setup_finetune.yaml` — not set (we only set batch_size, lr, lora_rank, epochs)
3. **Recipe defaults** ← torchtune's `8` wins here
4. Argparse defaults — `1`

## How base_recipe got into experiments

PR #260 (Jan 18, 2026) added the `base_recipe` feature and the design-experiment skill's `param_selection.md` actively recommends it:

> **Yes (recommended):** Specify `base_recipe` (e.g., `"llama3_2/1B_lora_single_device"`) and only set parameters you're actively varying.

The scaffold-torchtune agent then copies `base_recipe` into `setup_finetune.yaml`, where it pulls in torchtune's upstream defaults for any parameter we don't explicitly override.

## The fundamental issue

`base_recipe` conflates two things:
- **Model architecture** (which model component to use, tokenizer path, etc.) — useful
- **Training hyperparameters** (gradient accumulation, learning rate, etc.) — dangerous when they silently override our defaults

Our `finetune_template.yaml` is already our custom default config. Having `base_recipe` as an alternative source of defaults creates a silent override path that's hard to debug.

## Suggested fix

1. Stop recommending `base_recipe` in design-experiment skill docs
2. Stop including `base_recipe` in scaffolded `setup_finetune.yaml` by default
3. The `config_recipe_loader.py` code can remain for advanced users who explicitly want recipe inheritance

## Affected experiments

- `acs_income_bs32_2026-02-23` — gradient_accumulation_steps was 8 instead of 1
- Likely all experiments since Jan 18 that used `base_recipe` without explicitly setting gradient_accumulation_steps

cc @niznik-dev @sarahepedersen

niznik-dev:
Closes #332

## Summary

**base_recipe removal:**
- Remove `base_recipe` from design-experiment skill's "Step 4: Design Training Runs" — no longer asks users if they want recipe defaults
- Remove `base_recipe` field from `experiment_summary.yaml` template
- Remove `base_recipe` parsing/generation from `scaffold-torchtune.md` agent docs
- Remove `base_recipe` verification from `experiment_generation.md`
- Fix `gradient_accumulation_steps` default documentation from "1 (or 8 from recipes)" to "1"

**Stale baseline cleanup:**
- Remove the `baseline` section from the `experiment_summary.yaml` template — baselines now come from control runs, not arbitrary hardcoded numbers (leftover from before #297 removed baseline separation from reports)
- Remove the "Baseline for Comparison" step from `param_selection.md`
- Remove stale "Base recipe" reference from `validation.md` checklist

The `config_recipe_loader.py` code and `setup_finetune.py --base_recipe` CLI flag are preserved for advanced users who explicitly want recipe inheritance. The `workflow_test_recipe.yaml` integration test also remains to validate the feature still works.

## New Dependencies

None

## Testing Instructions

Run the LoRA Comparison workflow test to verify design → scaffold → run still works without `base_recipe`:
```
"test the workflow" → LoRA Comparison variant
```

Verify that scaffolded `finetune.yaml` uses `gradient_accumulation_steps: 1` (not 8).

### DIFF
--- a/.claude/skills/design-experiment/param_selection.md
+++ b/.claude/skills/design-experiment/param_selection.md
@@ -215,25 +215,6 @@ See `references/scorers.md` for the full list of available scorers, their parame
 
 **Important:** Base models evaluate once per task (no epoch suffix), fine-tuned models evaluate per epoch.
 
-### Baseline for Comparison
-
-**If no control run is included in the experiment**, ask: "What should results be compared against in the analysis report?"
-
-**Options:**
-- **Random chance** - For classification tasks, use 1/num_classes (e.g., 0.5 for binary)
-- **Known value from prior work** - Reference a specific accuracy from a previous experiment
-- **No baseline needed** - Exploratory experiment, will compare runs to each other
-
-**Record in experiment_summary.yaml:**
-```yaml
-evaluation:
-  baseline:
-    accuracy: 0.5
-    source: "random chance (binary classification)"
-```
-
-**Note:** If a control run (type: "control") is included, it automatically serves as the baseline and this section can be omitted.
-
 ### Create Evaluation Matrix
 
 Generate the evaluation matrix in experiment_summary.yaml:
--- a/.claude/skills/design-experiment/templates/experiment_summary.yaml
+++ b/.claude/skills/design-experiment/templates/experiment_summary.yaml
@@ -128,13 +128,3 @@ evaluation:
       epochs: [int]                # REQUIRED: List of epoch numbers (0-indexed)
                                    # Use null for control runs (no epochs)
 
-  baseline:
-    # OPTIONAL: Comparison point when no control run exists in experiment
-    # Use ONE of the following approaches:
-
-    # Approach 1: Known accuracy value (e.g., random chance)
-    accuracy: float                # Known baseline accuracy (e.g., 0.5 for binary classification)
-    source: string                 # Where this value comes from (e.g., "random chance (binary)")
-
-    # Approach 2: Explicit run as baseline (alternative to type: "control")
-    # run: string                  # Run name to use as baseline (matches runs[].name)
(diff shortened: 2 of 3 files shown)


niznik-dev:
@sarahepedersen Claude decided to nuke it - any potential downsides we're missing?

.claude/skills/design-experiment/param_selection.md
```diff
@@ -237,25 +215,6 @@ See `references/scorers.md` for the full list of available scorers, their parame
 
 **Important:** Base models evaluate once per task (no epoch suffix), fine-tuned models evaluate per epoch.
 
-### Baseline for Comparison
```
sarahepedersen: is this PR also removing the baseline comparisons for analysis in cases where no model is run? just want to make sure that is documented in the description

niznik-dev: OK updated. Thanks for asking!

PR #410, opened 2026-04-02
Fix evaluation template for multiple GPUs

sarahepedersen:
Closes #{ISSUE_NUMBER}

# Description

- When offline, the HF provider defaults to `"cuda:0"` (single GPU) for the device_map, which causes OOM on models that require spreading across GPUs. 
- This consults the model --> GPU mapping in `model_configs.py` and passes through `device="auto"` to distribute across available GPUs when more than one is required for the model. 

## New Dependencies

NA


niznik-dev (approved):
Looks good! Thanks for fixing this.

PR #414, opened 2026-04-05
Add tabular-to-text workflow

### LINKED ISSUE #369: Create tabular-to-book skill for converting structured data to prose format
## Summary

Create a Claude Code skill that converts tabular/structured data into natural language prose ("book" format) suitable for LLM consumption. This is the format used by folktexts-style tasks where a row like `{age: 45, education: Bachelor's, income: 75000}` becomes a narrative description like "This person is a 45-year-old with a Bachelor's degree who earns $75,000 per year."

## Motivation

Multiple potential users have asked for this capability. Currently, converting tabular data to prose is done ad hoc per experiment. A reusable skill would make this a standard part of the cruijff_kit workflow and lower the barrier for new users working with survey, census, or other structured datasets.

## Requirements

- Should support arbitrary tabular data (CSV, parquet, etc.), not just ACS/folktexts
- Implemented as a Claude Code skill (interactive, guided workflow)
- User should be able to configure which columns to include, how they're described, and the prose template/style
- Output should be in a format compatible with existing fine-tuning and evaluation pipelines

## Beta Readiness

This is item #1 on the [Beta Readiness](https://github.com/niznik-dev/cruijff_kit/wiki/Beta-Readiness) checklist (New Capabilities).

sarahepedersen:
Closes #{369}

# Description

Adds a tabular-to-text pipeline for tabular data (csv, parquet, dta) in cruijff_kit. Core additions are the `schema.yaml` and the `data_generation` block in the `experiment_summary.yaml` - those specifications are integrated downstream through the `text_gen` module into the `convert-tabular-to-text` skill. Enables the user to vary featurization, prompting formatting (dictionary vs narrative), and programmatically add perturbations (swapping out synonyms, shuffling column order, phrasing, etc.) between training and evaluation. 

See the `TEXT_GEN.md` file for full description of the (1) new skills workflow and (2) the created Python module - written by me without Claude while iterating on this one (in case you notice inferior prose...)

## New Dependencies

`jinja2` for the narrative templating , optionally `anthropic` for LLM-generated row-by-row prompts 

# Testing Instructions

tests for Python module text_gen in `tests/unit/text_gen`

sample experiment in my scratch dir under `acs_dict_to_narr_2026-04-05`, sample generated datasets/metadata, Jinja2 templates, and schema in `ck-data`


text_gen/lib/templates/narrative.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+        rendered = self._template.render(features=feat_dicts).strip()
+
+        # Choose splitting strategy based on whether the template uses
+        # explicit ||| delimiters or relies on period-splitting.
+        if "|||" in rendered:
+            chunks = [c.strip() for c in rendered.split("|||") if c.strip()]
+        else:
+            # Legacy: split on "." — one sentence per feature.
```
niznik-dev: I'm not sure what a sane default is for this, but '.' is going to cause some issues:

Abbreviations: "Dr. Smith"
Decimal numbers: "earned $52,000.50"
Ellipses, URLs, acronyms with periods ("U.S. citizen")

text_gen/lib/templates/llm_narrative.py
```diff
@@ -0,0 +1,173 @@
...
+
+        if cache_key in self._cache:
+            text = self._cache[cache_key]
+        else:
+            text = self._call_api(features)
+            self._cache[cache_key] = text
+            self._save_cache()
+
+        # Split into sentences, one Segment per sentence
+        sentences = [s.strip() for s in text.split(".") if s.strip()]
```
niznik-dev: Some comment re: '.' as default delimiter

tabular_to_text_gen/lib/templates/llm_narrative.py
```diff
@@ -0,0 +1,173 @@
...
+            prompt += f"\n\nAdditional instructions: {self._style_guidance}"
+
+        response = self._client.messages.create(
+            model=self.MODEL,
+            max_tokens=300,
+            messages=[{"role": "user", "content": prompt}],
+        )
+        return response.content[0].text.strip()
+
+    def _make_cache_key(self, features: list[tuple[ColumnSchema, str]]) -> str:
```
niznik-dev, line 133: This seems like it would benefit from the model name also being included to avoid caching issues if the model changes - MxC

text_gen/lib/templates/dictionary.py
```diff
@@ -0,0 +1,54 @@
...
+                        "synonyms": col_schema.synonyms,
+                        "shorthand_map": col_schema.shorthand_map,
+                        "restatements": col_schema.restatements,
+                    },
+                )
+            )
+        return segments
+
+    @staticmethod
+    def _format_value(raw_value: str, col_schema: ColumnSchema) -> str:
```
niznik-dev: This is duplicated in all the children and should live in the base template instead - MxC

text_gen/convert.py
```diff
@@ -0,0 +1,375 @@
...
+
+    return df.iloc[selected].reset_index(drop=True)
+
+
+def load_condition_from_file(conditions_file: str, condition_name: str) -> dict:
+    """Load a single condition's config from a conditions YAML file."""
+    with open(conditions_file) as f:
+        conditions = yaml.safe_load(f)
+
+    if condition_name not in conditions.get("conditions", conditions):
```
niznik-dev: Will we really support both of this possibilities? We should be opinionated about the format unless there's a strong reason not to be. -MxC

text_gen/lib/acs_pums_value_maps.py
niznik-dev, line 1: [NB] I'm not crazy about this being a Python file; seems like it should be yaml or something. But I won't die on this hill (hence non-breaking)

text_gen/lib/features.py
```diff
@@ -0,0 +1,88 @@
...
+        logger.warning(msg)
+
+    return warnings
+
+
+def _is_missing(raw_value) -> bool:
+    """Return True if raw_value represents a missing/NaN value."""
+    if raw_value is None:
+        return True
+    if isinstance(raw_value, float) and raw_value != raw_value:
```
niznik-dev: Let's use math.isnan() here instead for readability

text_gen/ARCHITECTURE.md
```diff
@@ -0,0 +1,351 @@
...
+
+Examples:
+- `dict_full_train_s42.json` + `dict_full_train_s42.meta.json`
+- `dict_synonym_test_s42.json` + `dict_synonym_test_s42.meta.json`
+
+---
+
+## Test Suite
+
+Tests live in `text_gen/tests/` and run with pytest:
```
niznik-dev: This never got updated; they're in `tests/unit/text_gen/` -MxC

text_gen/convert.py
niznik-dev: We should also change the pattern to match lines 217-218 (store the get in data and then use it in the if statement)

text_gen/lib/perturbations/synonym.py
```diff
@@ -0,0 +1,41 @@
...
+    result = []
+    for seg in segments:
+        synonyms = seg.metadata.get("synonyms", [])
+        if len(synonyms) < 2:
+            result.append(seg)
+            continue
+
+        new_name = rng.choice(synonyms)
+        # Re-render text with the new display name
+        new_text = f"The {new_name} is: {seg.value}."
```
niznik-dev: This is a problem if we weren't using dictionary in the first place - should think through that

text_gen/lib/perturbations/shorthand.py
```diff
@@ -0,0 +1,53 @@
...
+        else:
+            # Invert the map: abbreviated -> full
+            lookup = {v: k for k, v in shorthand_map.items()}
+
+        new_value = lookup.get(seg.value, seg.value)
+        if new_value == seg.value:
+            result.append(seg)
+            continue
+
+        new_text = f"The {seg.display_name} is: {new_value}."
```
niznik-dev: Same comment about only working with dictionary template

niznik-dev (changes requested):
Here's a first round of changes that we should work on - wanted to get these out there while I keep looking things over and run some experiments

text_gen/lib/templates/narrative.py
sarahepedersen: fixed to not support periods anymore

text_gen/convert.py
sarahepedersen: requiring that it be nested since that is what is in the fixture and generated by `experiment_summary.yaml`

text_gen/lib/acs_pums_value_maps.py
sarahepedersen: whoops I didn't mean to put this one in the repo in the first place -- it should exist in the sample schema (i'll add)

text_gen/lib/perturbations/synonym.py
sarahepedersen: replaced with regex

niznik-dev:
Heads up, Sarah — PR #423 collision survey, filed preemptively.

**Situation report.**

Branch `418-model-organisms` introduces an optional declarative block in `experiment_summary.yaml` and a dispatcher at `tools/experiment/prepare_data.py` that generates datasets before `scaffold-experiment` launches its subagents. Current shape:

```yaml
data_generation:
  model_organism:
    - name: bits_parity_k10_memo
      input_type: bits
      rule: parity
      k: 10
      ...
    - name: bits_first_k8_memo    # multi-dataset per experiment is supported
      ...
```

**Disclosure, for the record.** I skimmed this PR around the time I was drafting chunks 5 and 6 of 418. It is entirely possible that "put data-generation behind a declarative block" is a pattern I absorbed from you rather than one I independently synthesized. If the converged shape ends up closer to yours than to mine, that is consistent with the chronology.

**Collision surface.**

Your shape, as read from this PR:

```yaml
data:
  training: { ... }
  data_generation:
    tool: tabular_to_text_gen
    source: ...
    conditions:
      condition_a: { features: [...], template: ... }
      condition_b: { ... }
```

418's shape:

```yaml
data:
  training: { ... }
data_generation:
  model_organism:
    - { name: dataset_a, rule: parity, k: 10, ... }
    - { name: dataset_b, rule: first,  k: 8,  ... }
```

Three structural disagreements:

- **Where it lives** — yours is nested under `data:`; mine is top-level.
- **How many** — yours holds exactly one generator invocation per experiment (multiplicity lives inside `conditions:`); mine holds a list of standalone invocations.
- **How the tool is identified** — yours reads a `tool:` field; mine reads the YAML key name.

Each shape is internally consistent. Neither accommodates the other use case without modification.

**Reconciliation options, enumerated without preference:**

1. Adopt your shape verbatim. 418 drops multi-dataset support and compensates at the skill layer.
2. Hybrid: `data.data_generation:` as a list of entries, each tagged with `tool:`. Your case is a 1-element list; 418's is N.
3. Adopt 418's shape. Your block becomes another key under the top-level `data_generation:`.
4. Maintain two separately named blocks. Semantic divergence as a feature, not a bug.

**Status.**

PR #423 is ready for review. Disposition options: hold indefinitely, merge now and rebase later, or migrate to the agreed shape before merging. Sequencing is yours to set.

Awaiting input.

—MxC

niznik-dev:
**Follow-up: plan set, awaiting dissent.**

Your Slack note yesterday resolved the shape question — nest under `data:`, dispatch on a `tool:` field. Agreed.

Separate internal decision on 418's side: **we will not ship multi-dataset support in this PR.** It was speculative capability that no current workflow needs. Deferring it collapses the remaining "list vs dict" disagreement and lets 418 match your shape exactly (single dict, one tool, one dataset per experiment).

**Concrete migration plan for 418:**

1. Move `data_generation:` under `data:`, single dict, with an explicit `tool: model_organism` field — verbatim your proposed shape.
2. Rewrite the three reference YAML examples in `.claude/skills/design-experiment/references/model_organisms.md`.
3. Simplify `tools/experiment/prepare_data.py` to dispatch on `tool:` without the multi-entry loop.
4. File a follow-up issue for re-adding multi-dataset later, with the list-of-dicts vs keyed-multimap design decision explicitly scoped out.
5. Land as chunk 7 on the 418 branch: "Align data_generation to #414 shape; defer multi-dataset to follow-up."

**Merge consequences.** Post-migration, test-merging #414 into 418 reduces to **one trivial conflict**: both branches add entries to the `packages = [...]` list in `pyproject.toml`. ~30 seconds of hand-merge, no semantic decision required.

Starting the migration now. If you object to any part of this — especially the `tool: model_organism` name or the multi-dataset deferral — ping and we'll halt and revert.

—MxC

niznik-dev:
Delta review: prior blockers all cleanly fixed — thanks Sarah.

One non-blocking flag for a follow-up on our side: `tools/inspect/report_generator.py:567-580` silently reports 0 validation samples under the new bundled layout. The train sidecar now carries validation in `extra_splits` (`tabular_to_text_gen/lib/output.py:201`), but `_read_splits_from_metadata` still looks for a sibling `_validation_{hash}.meta.json` file and gives up quietly when absent. No reporting test exercises `validation_ratio`, so CI is green. Fix: read `extra_splits` before the sibling-file loop.

—MxC

niznik-dev:
To clarify the plan: I'll push the fix to this branch shortly. Once CI is green again, we can merge.

—MxC

niznik-dev (approved):
This is going to be so cool thanks for working on this!

Issue #457, opened 2026-05-05
Evaluate splitting tool-agnostic env config from claude.local.md


niznik-dev:
## Context

In [#365](https://github.com/niznik-dev/cruijff_kit/issues/365), @msalganik raised:

> Should hardware and directory information go in `claude.local.md` or other file. I'd prefer to have something that is not claude specific.

While building the `setup` skill (#365) we deferred this question — the skill ships writing to `claude.local.md` as the single config location, matching the status quo. This issue is the placeholder for revisiting the split.

## What needs to be decided

Whether to split the current `claude.local.md` into:

- A **tool-agnostic** env config file (HPC paths, SLURM defaults, conda env, model locations, wandb config) — usable by anyone, including users following the manual `WORKFLOW_GUIDE.md` who never touch Claude
- A **Claude-specific** preferences file (Claude Code session behavior, MxC signatures, anything genuinely Claude-only)

## Plausible motivations for the split

These are inferred — Matt's comment didn't spell out the reasoning. Worth Matt confirming or adding before this gets scoped.

| Motivation | Strength |
|---|---|
| **Tool independence** — `WORKFLOW_GUIDE.md` documents manual (no-Claude) workflows; making non-Claude users populate a file named `claude.local.md` is misleading | Strongest case |
| **Future-proofing** — if codex.local / aider.local files ever exist, env config gets duplicated | Speculative; not a current pain point |
| **Discoverability** — a new user grepping for "where do I configure HPC paths?" wouldn't think to open a Claude-specific file | Weak — README + `/setup` skill addresses this |

## Blast radius

`claude.local.md` is currently referenced from 15 files:

```
.claude/skills/analyze-experiment/SKILL.md
.claude/skills/convert-tabular-to-text/SKILL.md
.claude/skills/design-experiment/{SKILL.md, experiment_generation.md, param_selection.md, validation.md, templates/experiment_summary.yaml}
.claude/skills/scaffold-experiment/{SKILL.md, logging.md, evaluators/inspect_agent.md, optimizers/torchtune_agent.md}
.claude/skills/summarize-experiment/SKILL.md
docs/{ACS_EXAMPLE.md, ARCHITECTURE.md}
README.md
```

A split would touch all of these plus migrate every existing user (each has a `claude.local.md` they'd need to bisect).

## Acceptance criteria for this issue

This is a **scoping** issue, not an implementation issue. Done when:

[...]

PR #454, opened 2026-05-03
Add create-quiz skill (#453)

msalganik:
Closes #453

# Description

Adds a `create-quiz` skill that turns a completed experiment into a self-contained, self-grading HTML quiz. The quiz tests a recipient's intuition about the *results* (not the experimental process); the full write-up with figures is hidden until the recipient clicks "See Final Score & Full Analysis", so it works as an intuition test before serving as a teaching artifact.

The HTML output is portable (inline CSS, vanilla JS, base64 images, no backend) and the page banner advertises a link to the cruijff_kit GitHub repo.

## Skill content

- `.claude/skills/create-quiz/SKILL.md` — top-level workflow, audience-and-assumption framing, validation contract.
- `.claude/skills/create-quiz/{parsing,question_catalog,generation,logging}.md` — modular sub-files following the existing experiment-skill pattern.
- `.claude/skills/create-quiz/evals/evals.json` — three skill-creator-style test prompts (single experiment, two-experiment comparison, prerequisite-missing branch).
- `.claude/skills/create-quiz/evals/iteration-1/{benchmark.json, feedback.md}` — first iteration's validator feedback (5 SKILL.md defects identified by an independent subagent and fixed in-iteration before this PR).

## Default question mix

Three multiple_choice + two numerical_estimate (percent form, 0–100, with `min`/`max` HTML5 attributes that reject out-of-range submissions before grading) + one ranking. `image_read`, `equation_or_baseline`, `free_text_intuition`, and operational `anomaly_hunt` are still supported in the renderer schema but explicitly excluded from the default mix (rationale documented in `question_catalog.md`).

## Renderer

- `src/tools/quiz/render_quiz.py` — Jinja2 + python-markdown. Pre-renders intro / prompts / explanations / full-report from markdown. Embeds PNG references in the report markdown as base64 data URIs. Strips local filesystem paths from the embedded JSON so view-source doesn't leak `/scratch/...` paths.
- `src/tools/quiz/templates/quiz.html.j2` — single-file output: brand banner with cruijff_kit GitHub link, title, intro, "Full experimental specification" `<details>` panel (the verbatim experiment_summary.yaml), questions, sticky score bar with the "See Final Score & Full Analysis" button, score panel, hidden writeup `<details>` panel that's revealed and auto-opened on submit.

## New Dependencies

- `markdown` (python-markdown) — runtime dependency for the renderer.

# Testing Instructions

Unit tests:

```bash
[...]

PR #458, opened 2026-05-05
Add /setup skill and trim design-experiment prereq inline check

### LINKED ISSUE #365: Create onboarding skill for new users
## Summary

Create a Claude Code skill (e.g., `onboard` or `setup`) that guides new users through initial configuration so they're ready to run their first experiment. Should trigger naturally when a user is new or when `claude.local.md` doesn't exist.

## Motivation

Getting started with cruijff_kit requires configuring several environment-specific settings (HPC paths, conda environments, SLURM defaults, model locations, etc.). Currently this means reading multiple docs and manually creating `claude.local.md` from the template. An interactive onboarding skill would make this frictionless.

### Detection
- Could be invoked explicitly (`/onboard`)
- Could also be suggested proactively if `claude.local.md` is missing when the user tries to run an experiment

### Walkthrough (incremental, with explanations)

1. **Welcome & orientation** — Brief overview of what cruijff_kit does and what we're about to set up
2. **`claude.local.md` creation** — Walk through the template interactively:
   - HPC username and scratch directory
   - SLURM defaults (partition, GPU type, account)
   - Conda environment name
   - Model cache / HuggingFace home location
3. **Environment validation** — Verify the basics work:
   - Can we find the conda env?
   - Do the scratch paths exist?
   - Is `gh` accessible (with the conda prefix)?
   - Are required modules available?
4. **Prerequisites check** — Quick scan against PREREQUISITES.md:
   - HuggingFace account & token
   - Model access (e.g., Llama gated models)
   - Any missing Python packages
5. **Ready to go** — Point them to their first experiment (link to README workflow or suggest `design-experiment`)

Each step should explain *why* the setting matters, not just ask for a value. User can skip steps they've already handled.

## Notes

- Use `claude.local.md.template` as the source of truth for what needs configuring
- Don't overwrite an existing `claude.local.md` — if one exists, offer to review/validate it instead
- Keep the tone welcoming — this is someone's first impression of the project
- Consider: should this also cover git workflow basics (branch naming, PR format) or keep it purely technical setup?

niznik-dev:
Closes #365

# Description

Adds a `/setup` skill — the new first-touch experience for cruijff_kit users — and trims the inline `claude.local.md` prerequisite check from `design-experiment` so the gating responsibility lives in one place.

## What `/setup` does

Two mutually-exclusive modes, picked based on whether `claude.local.md` exists:

- **Greenfield** — walks `claude.local.md.template` section-by-section, with a one-line rationale per section, auto-detecting `whoami` / `id -gn` / `pwd` (treating those as defaults to confirm, not facts to write — group case can differ from filesystem path case on some clusters). Allows `[skip]` for optional sections (W&B, shared datasets). Confirms before writing. Never overwrites without an explicit "start over" instruction.
- **Validate** — read-only health check on an existing `claude.local.md`. Three classes:
  - Placeholder scan in settings sections only (skips `## Quick Commands` by design — the `<placeholder>` tokens in example commands are intentional).
  - Required-field presence (`Cluster`, `Username`, `Group`, `Scratch directory`, SLURM `Account`, `Default conda environment`).
  - Lightweight env probes — conda env exists (gated on the configured anaconda module if needed), scratch dir writable, `gh` on PATH after activation, module system available if the file calls for it.

Returns a structured GREEN / YELLOW / RED report with concrete remediation per fail. Partial-pass is normal and useful (e.g. `gh` not yet installed in the env shouldn't block you from running `/design-experiment`).

A "start over" path renames the existing `claude.local.md` to a dated `.bak` (with collision counter) before re-entering greenfield mode — users never lose their prior config to a destructive command.

## What changes elsewhere

- `design-experiment/SKILL.md:14-53` — the inline "does `claude.local.md` exist? are placeholders filled in?" gate (~40 lines) is replaced with a one-paragraph pointer to `/setup`. The inline check duplicated work and couldn't grow into validate-mode without making `design-experiment` substantially longer.
- `CLAUDE.md` — new "Setup" subsection in the skills list, before "Primary Workflows".
- `README.md` — Configuration section now recommends `/setup` before the manual `cp claude.local.md.template` route.

## Skill development methodology

Skill content was iterated against a [skill-creator](https://github.com/anthropics/skills/tree/main/skills/skill-creator)-style validator subagent that audited `SKILL.md` / `walkthrough.md` / `validation.md` / `logging.md` against the 4 prompts in `evals/evals.json`. Seven defects found and fixed in-iteration before this PR:

1. Missing `RENAME_BACKUP` action / start-over procedure (eval #4 was uncovered)
2. Frontmatter trigger phrases didn't cover several eval prompts ("validate my setup", "is my environment configured correctly?", "health check")
3. `id -gn` group case can differ from filesystem path case on some clusters (would silently produce a wrong scratch path in greenfield mode)
4. Conda probe assumed `conda` was on PATH — needs a `module load <anaconda_module>` preamble on clusters that gate it
5. Hardcoded `anaconda3/2025.6` (Della-as-of-this-month) in skill body — should be read from the config, not assumed
6. Placeholder-scan awk script had a section-boundary bug — replaced with a simpler "stop reading at `## Quick Commands`" approach
7. Over-promised auto-detect list (`which conda` was listed but never used downstream)

Per the lesson from PR #454, the validator's working files (`evals/iteration-1/`) are NOT included — those were the dev-loop scaffolding, not something a future contributor needs to wade through.

## Deferred / out of scope

- Splitting `claude.local.md` into a tool-agnostic env config + Claude-specific preferences (raised by @msalganik on #365): tracked separately as **#457** on milestone 0.3.1. This PR keeps everything in `claude.local.md` for now to avoid a wide-blast-radius migration in the same change.
- Validating HuggingFace tokens or other authenticated external services — point at `docs/PREREQUISITES.md` instead.
- Auto-installing the conda env — `/setup` verifies the env exists; users still run `make install` themselves.

## New Dependencies

None.

# Testing Instructions

This is documentation-only skill content; there are no unit tests to run. Manual smoke-test plan against the four eval prompts in `.claude/skills/setup/evals/evals.json`:

1. **Eval 1 — greenfield**: in a fresh clone with no `claude.local.md`, invoke `/setup` with "I just cloned cruijff_kit and I have no idea what to configure first." Expect: section-by-section walkthrough with rationale, auto-detect of whoami/id/pwd as defaults to confirm, `[skip]` honored for W&B, `claude.local.md` written only after explicit yes.
2. **Eval 2 — validate clean**: with a fully-filled `claude.local.md` and the configured conda env actually present, invoke `/setup` with "is my environment configured correctly?". Expect: GREEN summary, no overwrite, pointer to `/design-experiment`.
3. **Eval 3 — validate with placeholders**: leave 2+ unfilled placeholders in settings sections (e.g. `<your_slurm_account>`). Invoke `/setup` with "validate my setup". Expect: per-line YELLOW/RED findings with line numbers + section + concrete fix, no false positives from `<placeholder>` tokens in `## Quick Commands`.
4. **Eval 4 — start over**: with an existing `claude.local.md`, invoke `/setup` with "I want to start over." Expect: confirmation prompt, rename to `claude.local.md.bak.<YYYY-MM-DD>`, then greenfield walkthrough.

Lint: no Python or YAML files added.


.claude/skills/ck-setup/SKILL.md
niznik-dev, line 1: I question whether this should be called `/setup` because it may conflict with the Claude namespace and be confusing

niznik-dev:
Otherwise my human readthrough revealed nothing obviously bad. Going to test some more before merging.

niznik-dev:
The skill description is also far too long - let's shorten that

niznik-dev:
When the skill loads it should set expectations as well as how long it will take

niznik-dev:
Greenfield Question 2 should be more straightforward about where should output be stored

niznik-dev:
The skill should ask the user if they have an existing slurm script for referencing

PR #441, opened 2026-04-24
Folder reorganization: src/ layout, blueprints/, ck-projects/ unified scratch

niznik-dev:
Closes #437

# Description

Executes the atomic folder reorg committed in `edi-ccore/reviews/folder_reorg_plan_2026-04-22.md`. Deviations from the plan are documented in the addendum at the bottom of that same doc.

**What changed (15 commits, ~40% of tracked repo files touched):**

### `src/` layout (chunk 2)
- Library code moved under `src/` (flat layout): `src/tools/`, `src/utils/`, `src/tabular_to_text_gen/`
- `pyproject.toml`: `package-dir = {cruijff_kit = "src"}` — imports remain `from cruijff_kit.X import Y`
- `tabular_to_text_gen.*` imports rewritten to `cruijff_kit.tabular_to_text_gen.*` across tree

### `blueprints/` (chunks 4 + F)
- 5-slot shape (README, inspect_task.py, generate_data.py, optional modifiers/ + baseline.py)
- `blueprints/capitalization/` + `blueprints/folktexts/` built from `experiments/*/` content
- `blueprints/model_organism/`: README-only pointer; framework code stays in `src/tools/model_organisms/` (see addendum)
- `experiments/` directory retired entirely
- `synthetic_twins` retired — superseded by model_organisms framework
- **Chunk F:** initially landed as `projects/`, then renamed to `blueprints/` to disambiguate from the `ck-projects/{project}/` runtime scratch concept (group meeting decision, 2026-04-27). YAML schema field `experiment.project` stays as-is — the lookup pattern is now `blueprints/{experiment.project}/`.

### `src/tools/model_organisms/` (chunk 3)
- `sanity_checks/model_organisms/` → `src/tools/model_organisms/` (peer of torchtune, inspect, slurm, experiment)
- `sanity_checks/` umbrella retired (other 5 subdirs already deleted in #434)

### Unified `ck-projects/` scratch (chunk 5)
- `ck-experiments/` + `ck-outputs/` + `ck-sanity-checks/` → `ck-projects/{project}/{experiment_name}/`
- Configs, checkpoints, logs, eval outputs now nest inside each experiment dir per run
- `experiment.project` field added (REQUIRED); `experiment.type` / `metadata.type` removed
- `claude.local.md.template`, test fixtures, skill docs, workflow test specs all updated

### Parquet workflow support removed (chunk 1)
- Dropped from `setup_finetune.py` fine-tuning path (unused for months)
[...]

Issue #372, opened 2026-03-13
Audit and standardize naming conventions across the codebase


niznik-dev:
## Summary

Review and standardize naming conventions across cruijff_kit — file names, function names, config keys, YAML fields, etc. The goal is consistency so that users and contributors can predict naming patterns rather than memorize them.

## Possible areas to look at

- Config key naming (snake_case? camelCase? mixed?)
- File naming patterns across tools/ and skills/
- Function and variable naming consistency
- YAML field names in experiment_summary.yaml and other configs

## Next steps

@sarahepedersen and @msalganik — this came up in the 3/13 group discussion and it sounded like you two had more specific ideas about what needs standardizing. Please add your thoughts here so we can scope this properly.

## Beta Readiness

This is item #7 on the [Beta Readiness](https://github.com/niznik-dev/cruijff_kit/wiki/Beta-Readiness) checklist (Cleanup).

niznik-dev:
After merging the folder reorg (#437) early next week, I'll take a short triage pass on this — mainly spotting anything *urgent* to rename before v0.3 ships, vs. items that can wait for a 0.3.x patch. Not the full audit. Will report back here with a candidate list before any renames.

—MxC

niznik-dev:
<img src="https://raw.githubusercontent.com/niznik-dev/llm-personas/main/fusions/mattie/mattie-icon.png" width="36" align="absmiddle"> 🎻 **MxC** ✳️

cc @msalganik @sarahepedersen — picking this umbrella back up.

We ran a naming audit across three dimensions — YAML/config keys, Python file & function names, and directory/skill/artifact names — to give this umbrella issue a concrete, bounded shape. This first comment lists the changes we're confident enough to **just do**; the follow-up comments each raise **one open decision** for you two.

### Two principles that bound the audit

1. **Ours vs. external contract.** We can rename our own schema, files, and folders. We **cannot** rename keys that are torchtune's or inspect-ai's contract — `lr`, `seed`, `batch_size`, `epochs`, `max_seq_len`, `output_dir`, the `@task`/`@scorer`/`@metric` names, the `finetune.yaml` recipe keys. Roughly half the raw "inconsistencies" are inherited from those tools and are off-limits; we've excluded them.
2. **Rename vs. document.** Some fixes are renames; others are just writing down a latent convention so future drift stops. We've split accordingly.

And one process rule: **each high-reference rename ships as its own PR**, never bundled, so a regression is easy to bisect and revert.

[...]

PR #349, opened 2026-03-06
Add compute estimates to design-experiment based on previous runs

sarahepedersen:
Closes #{243}

# Description

- Structures the output of `compute_metrics.json` in the `analyze-experiment` skill
- In `design-experiment`, Claude will search experiment directories (or user can direct Claude to specific experiments) for `compute_metrics.json` -- it will scale compute estimates with respect to changes in model size, batch size, LoRA rank, etc. Will also recommend batch size changes with respect to issue #{254}
- Adds `compute` section to the `runs` matrix in `experiment_summary.yaml` that specifies the estimated slurm time, number of gpus, and memory for each fine-tuning/evaluation job; utilized downstream by scaffolding to override defaults in the setup scripts 

## New Dependencies

NA

# Testing Instructions

need to test this more (leaving as draft PR for now)

PR #462, opened 2026-05-06
Add weighted_sum and weighted_sum_binary rules to model_organisms

### LINKED ISSUE #460: Add weighted_sum and weighted_sum_binary rules to model_organisms framework
## Motivation

The model_organisms framework (`src/tools/model_organisms/`) currently supports rules whose output is a single token from a small discrete space (`parity`, `majority`, `first`, `last`, `nth`, `length`, `constant`, `coin`, `min`, `max`). It does not support rules where the output is an arithmetic function of the input.

`weighted_sum` and `weighted_sum_binary` add the linear-DGP shape, letting downstream researchers ask questions like "Does fine-tuning recover a known linear data-generating process? How does the answer depend on sparsity, noise level, model size, or sequence length?"

This issue is for **library/infra work**: adding the rules and supporting framework changes. Specific experiments using these rules are out of scope.

## New rules

Two rules in `src/tools/model_organisms/rules.py`, both `applicable={"bits", "digits"}`:

- **`weighted_sum`** — returns the integer `w·x + intercept`, formatted as a spaced signed zero-padded string.
- **`weighted_sum_binary`** — returns `"0"` or `"1"`. Deterministic threshold when `noise_scale == 0`; Bernoulli sample with `p = σ((w·x + intercept) / noise_scale)` when `noise_scale > 0`. Stochastic samples are deterministic per `(weight_seed, sequence)` — note this differs from `coin`, which keys on the dataset `seed`. Keying label noise on the DGP seed is deliberate (see "Seed semantics" below).

Using these rules with `letters` raises a clear error.

Both rules are also restricted to `design: memorization` and `design: in_distribution`. Using them with `design: ood` raises a clear error — see *Out of scope* below for the rationale.

## `rule_kwargs` schema

| Field | Type | Default | Description |
|---|---|---|---|
| `weights` | `list[int] \| None` | `None` | Explicit weight vector of length `k`. If given, distribution params are ignored. |
| `weight_max` | `int` | `3` | Weights drawn uniformly from `{-W, …, -1, 1, …, W}` (excluding zero). |
| `sparsity` | `float` | `0.0` | Bernoulli probability that a drawn weight is then masked to zero. Applied after the magnitude draw. |
| `weight_seed` | `int \| None` | `None` (uses dataset seed) | Seed for the **DGP**: weight draws, sparsity mask, and (for the binary rule) the per-sequence label-noise stream. See "Seed semantics" below. |
| `intercept` | `int \| "balanced"` | `0` | DGP intercept. `"balanced"` resolves to `round(-sum(w_i) · E[x_i])` over the input alphabet (4.5 for digits, 0.5 for bits). |
| `noise_scale` | `float` | `0.0` | Binary rule only. `0` = deterministic threshold; `> 0` = Bernoulli sample. |

## Seed semantics

These rules expose **two seeds** with separate roles, so that DGP definition can be decoupled from sequence sampling:

- **`seed`** (the existing positional argument to `generate()`): controls *which sequences you observe* — primary sequence sampling, OOD sequence sampling (when applicable), the dataset-seed-keyed `coin` rule, and the bayes_accuracy MC sampler. Same role as for every other rule.
- **`weight_seed`** (in `rule_kwargs`, defaults to `seed` when `None`): controls *what the DGP is and how its noise behaves* — magnitude draws, the sparsity mask, and (for `weighted_sum_binary` with `noise_scale > 0`) the per-sequence Bernoulli stream.

Useful patterns this enables:
- **Same DGP, fresh sequences** — fix `weight_seed`, vary `seed`. Resolved weights, intercept, format width, and (where present) the per-sequence label-noise pattern stay constant; the train/val sample changes. Useful for studying sample-size effects or for averaging FT outcomes over sampling noise.
- **Same sample, different DGP** — fix `seed`, vary `weight_seed`. Sequence sampling stays fixed; the DGP changes. Useful for varying weight sparsity / magnitude while controlling for which sequences land in train vs. val.

Note the deliberate difference from `coin`: `coin` keys label draws on `(seed, sequence)`, so changing `seed` reshuffles labels. `weighted_sum_binary` keys label draws on `(weight_seed, sequence)`, so changing `seed` does **not** reshuffle labels — only `weight_seed` does. This is what makes "same DGP, fresh sequences" a clean comparison.

## Output format

Spaced + signed + zero-padded integer string. Format width is computed once per dataset from the resolved weights and intercept: `digits = ceil(log10(sum(|w_i|) · alphabet_max + |intercept| + 1))`, plus a sign token.

Example: `k=8`, weights `[1, 2, -1, 3, 0, 1, -2, 1]`, digits alphabet → max output 99 → 2-digit width → output 42 renders as `+ 4 2`.

This generalizes across all major BPE/SentencePiece tokenizers (Llama-2/3, GPT-4, Claude, Mistral, Qwen) because whitespace is a hard tokenization boundary.

## Framework change: `Rule.prepare` hook

The current rule signature `fn(seq, **kwargs) -> str` is per-example. `weighted_sum` needs dataset-level state (resolved weights, format width, Bayes accuracy). Extend the `Rule` dataclass:

```python
@dataclass(frozen=True)
class Rule:
    name: str
    fn: Callable[..., str]
    applicable: frozenset[str]
    prepare: Callable[..., dict] | None = None
```

When non-None, `prepare(input_type, k, N, seed, rule_kwargs) -> dict` is called once at generation time and its return value is merged into `rule_kwargs` before per-example `fn` calls. `generate.py` is updated to invoke this hook. All existing rules keep `prepare=None` — no behavioral change.

## Dataset metadata

Top-level dataset metadata is extended for these rules to record:

- `resolved_weights: list[int]`
- `intercept: int`
- `noise_scale: float` (binary only)
- `format_width: int`
- `bayes_accuracy: float` (binary stochastic only) — analytical or Monte Carlo estimate over the input distribution

Downstream users get ground-truth weights for coefficient-recovery analyses and the optimal-model upper bound without recomputation.

## Out of scope (file as separate follow-up issues)

- ICL baseline infra (an `n_shot` parameter on `inspect_task.py`). Strongly recommended for any downstream experiment using these rules — without it, FT improvements are uninterpretable (Vacareanu 2024).
- Tokenizer sanity check at data generation. Generic feature, not specific to these rules.
- Continuous (non-integer) weights and float output formatting.
- Additional weight distributions beyond uniform-integer.
- **OOD design with `weighted_sum` / `weighted_sum_binary`.** The resolved weight vector, intercept, and format width are dataset-level state shared across all splits, which creates several footguns under `design: ood`:
  - per-split DGP overrides (e.g. different `weights` in `ood_tests[i].rule_kwargs`) would be silently ignored, since primary's `resolved_weights` propagate to every split;
  - OOD splits with `k ≠ k_primary` would surface as a per-row length-mismatch exception during dataset construction rather than as an upfront validation error;
  - OOD splits that change `input_type` (bits ↔ digits) would silently render outputs at a different width than primary, because the `+0Nd` formatter pads but doesn't truncate.

  Supporting OOD properly requires explicit design — re-drawing weights creates a different DGP (so it isn't an OOD test of the *same* DGP), padding with zeros makes an arbitrary extrapolation choice, and per-split format widths break the "stable output format across the dataset" guarantee. Until a concrete length-generalization or distribution-shift use case appears, `generate()` rejects `design: ood` for these rules at validation time. File a follow-up issue if a use case emerges.

## Acceptance criteria

- [ ] `weighted_sum` and `weighted_sum_binary` registered; returned by `list_rules()`
- [ ] `Rule.prepare` hook implemented and invoked from `generate.py`; existing rules unchanged
- [ ] Output format stable across all examples in a dataset
- [ ] Dataset metadata records resolved weights, intercept, noise_scale, format width, and Bayes accuracy (binary stochastic)
- [ ] `generate()` rejects `design: ood` for `weighted_sum` and `weighted_sum_binary` with a clear error message
- [ ] Unit tests cover: explicit weights, distribution-drawn weights (deterministic per seed), `sparsity=1.0` → all-zero weights, deterministic threshold, stochastic sample (deterministic per `(seed, sequence)`), `intercept="balanced"` produces approximately balanced binary classes, `applicable` rejects `letters`, `design: ood` rejected for both rules, edge cases (`k=1`, all-zero weights, `sparsity=0`), recorded `bayes_accuracy` matches empirical optimal-classifier accuracy within statistical tolerance
- [ ] Documentation updated: rule docstrings (note `majority` and `constant` as special cases), parameter table in `.claude/skills/design-experiment/references/model_organisms.md`, examples in `blueprints/model_organism/README.md`
- [ ] `ruff check` and `ruff format --check` pass on changed files

## References

- Vacareanu et al. (COLM 2024). "From Words to Numbers: Your LLM Is Secretly A Capable Regressor When Given In-Context Examples." [arxiv 2404.07544](https://arxiv.org/abs/2404.07544) — establishes that pretrained LLMs already do regression in-context, motivating the ICL-baseline follow-up.
- Singh & Strouse (2024). "Tokenization counts: the impact of tokenization on arithmetic in frontier LLMs." [arxiv 2402.14903](https://arxiv.org/abs/2402.14903) — anchors the spaced-format choice.

msalganik:
Closes #460

# Description

Adds two new rules to the model_organisms framework that produce a known linear-DGP target, plus the framework support they need:

**New rules** (both `applicable={"bits", "digits"}`):
- `weighted_sum` — returns `w·x + intercept` as a spaced signed zero-padded integer string (e.g. `+ 4 2`).
- `weighted_sum_binary` — returns `"0"` or `"1"`. Deterministic `z > 0` threshold when `noise_scale == 0`; σ-Bernoulli sample with `p = σ(z / noise_scale)` when `noise_scale > 0`. Stochastic samples are deterministic per `(weight_seed, sequence)`.

**Framework changes:**
- `Rule.prepare` hook (optional `Callable[..., dict] | None`, default `None`): called once per `generate()` call to resolve dataset-level state (resolved weights, format width, Bayes accuracy). Returned dict merges into `rule_kwargs` for fn calls. All existing rules keep `prepare=None` — no behavioral change.
- `Rule.supports_ood` flag (default `True`): rules with dataset-level state opt out of `design="ood"` by setting `False`. `generate()` rejects with a clear error before any work is done. `weighted_sum` and `weighted_sum_binary` use this.
- Top-level dataset metadata for these rules records `resolved_weights`, `intercept`, `format_width`, `weight_seed`, `noise_scale` (binary), and `bayes_accuracy` (binary stochastic — analytical when `alphabet^k ≤ 2^20`, else 20k-sample MC).

**Seed semantics** (intentional decoupling, documented in code + skill ref + issue):
- Dataset `seed` controls *which sequences you observe* (sampling, `coin` labels, MC samplers).
- Rule-kwargs `weight_seed` (defaults to `seed`) controls *what the DGP is and how its noise behaves* — weight magnitudes, sparsity mask, and binary-stochastic per-sequence Bernoulli stream.
- Fixing `weight_seed` and varying `seed` yields fresh sequence draws of the same DGP with the same per-sequence label-noise pattern. This deliberately differs from `coin`, which keys label noise on the dataset seed.

**OOD rejection rationale** (filed in #460): with a single resolved weight vector, intercept, and format width fixed at the dataset level, OOD splits would (a) silently override user DGP attempts in `ood_tests[i].rule_kwargs`, (b) surface k-mismatch as a per-row exception rather than upfront validation, and (c) silently render at a different width if `input_type` differs across splits. Rejected at validation time until a concrete length-generalization use case appears.

**Documentation:** rule docstrings note `majority` and `constant` as special cases of the linear DGP family. Skill reference (`.claude/skills/design-experiment/references/model_organisms.md`) and blueprint README (`blueprints/model_organism/README.md`) updated with the full `rule_kwargs` schema, output format, seed-semantics paragraph, and OOD restriction.

## New Dependencies

None. Only stdlib (`itertools`, `math`, `random`).

# Testing Instructions

Run the unit tests:

```bash
pytest tests/unit/test_model_organisms.py -v
```

35 new tests across `TestPrepareHook`, `TestWeightedSum`, and `TestWeightedSumBinary` cover every acceptance-criterion bullet from #460:

- explicit weights and arithmetic correctness
- distribution-drawn weights deterministic per `weight_seed`
- `weight_seed` defaults to dataset `seed`
- drawn weights exclude zero before sparsity mask
- `sparsity=1.0` zeros all weights
- `intercept="balanced"` for bits and digits
- format-width edge cases (all-zero weights, k=1)
- letters input rejected
- weight-vector length mismatch rejected
- invalid `intercept`, `noise_scale`, `sparsity`, `weight_max` rejected
- `design="ood"` rejected for both rules
- deterministic threshold (`z > 0` rule)
- stochastic samples deterministic per `(weight_seed, sequence)`, change when `weight_seed` changes
- `intercept="balanced"` produces approximately balanced binary classes (with σ-noise)
- `bayes_accuracy` recorded only when stochastic; matches empirical Bayes-optimal classifier accuracy on enumerated 256-bit sequence space within 0.10

Also exercises the CLI:

```bash
python -m cruijff_kit.tools.model_organisms.generate \
  --input_type digits --rule weighted_sum --k 5 --N 10 --seed 1 \
  --design memorization \
  --rule_kwargs '{"weight_max": 3, "weight_seed": 99}' \
  --output ws.json --output_dir /tmp
```

Lint:

```bash
ruff check src/tools/model_organisms/ tests/unit/test_model_organisms.py
ruff format --check src/tools/model_organisms/ tests/unit/test_model_organisms.py
```

Both clean. Existing tests in `tests/unit/test_model_organism_inspect_task.py` and `tests/unit/test_prepare_data.py` also pass (no regressions in adjacent code paths).


niznik-dev:
Hi Matt — pushed two commits to this branch (heads-up so nothing's a surprise):

**`39c5b78`** — Merged `main` to bring this branch up-to-date with the v0.3.0-bound merges (#463, #468, #469, #458, #445, #349, #456, #454). No conflicts.

**`eab9350`** — Bundled some related housekeeping that pairs naturally with this PR:
- Renamed `blueprints/model_organism/` → `blueprints/model_organisms/` for symmetry with `src/tools/model_organisms/` (covers the dir-mismatch friction issue from #372 triage)
- Dropped legacy "synthetic" and "sanity check" framing across the model-organisms docs and the `design-experiment` skill — replaced with "sequence-labeling tasks with known ground-truth rules" or just "experiment"
- Removed the vestigial `sanity_check vs research experiment` dichotomy in `design-experiment` — the `experiment.type` schema field was already removed in #441, but a few prose references and a log-example field lingered
- Added CHANGELOG entries for both the new rules and the rename/cleanup

Context: Mattie wants this in v0.3.0 alongside your rules. The rename made sense to bundle since this PR already touches `blueprints/model_organism/README.md`, and folding the related terminology cleanup keeps the breaking-change blast radius in one PR rather than two.

If any of the rephrasing feels off (or if you'd rather split the housekeeping into a separate follow-up), happy to revert — just say the word. The rename itself is the most consequential change, since it shifts `experiment.project: model_organism` → `model_organisms` in any existing user yamls.

—MxC

niznik-dev (approved):
I cleaned this up a bit and think it's in good shape now

msalganik:
@niznik-dev https://github.com/niznik-dev/cruijff_kit/pull/462#issuecomment-4399578314 looks good to me.  I like the idea of getting the renaming in now.

PR #463, opened 2026-05-06
Replace calibration flag with logprobs capability switch (#461)

### LINKED ISSUE #461: Replace calibration flag with logprobs capability switch in model_organism task
## Problem

The `model_organism` inspect task at `src/tools/model_organisms/inspect_task.py:30` has a single flag `calibration: bool` that does two distinct things:

1. Enables logprob capture (`GenerateConfig(logprobs=True, top_logprobs=20)`)
2. Auto-appends `risk_scorer` to the scorer list

The name reflects the second behavior (the calibration analysis workflow) but obscures the first (the underlying capability). A user wanting logprobs for non-calibration analysis — confidence estimation, selective prediction, AUC, distributional comparisons — cannot discover the feature from the API.

We have a confirmed case where Claude, with source access, could not surface this flag in response to a "how do I get logprobs?" question. Naming-as-documentation: if the assistant can't find it, a human reading a YAML has no chance.

## Proposed change

Drop `calibration` entirely. Replace with two primitives plus a scorer-driven inference:

- `logprobs: bool = False` — capability switch, name says what it does.
- `top_logprobs: int = 20` — exposes the previously hardcoded value. Multiclass / tokenization variants may need more.

New signature:

```python
def model_organism(
    data_path: str,
    config_path: str = "",
    split: str = "validation",
    temperature: float = 1e-7,
    max_tokens: int = 5,
    use_chat_template: bool = True,
    logprobs: bool = False,
    top_logprobs: int = 20,
    vis_label: str = "",
) -> Task:
```

Logprobs auto-enable when a scorer in the configured list declares it needs them. **Use a `requires_logprobs = True` attribute on the scorer factory, not name-matching.** This matters because `numeric_risk_scorer` does NOT need logprobs (it parses a float directly from text completion), so any name-based heuristic like "anything matching `*risk*`" would over-include and force unnecessary logprob capture on that path. Only `risk_scorer` declares the attribute.

**Conflict semantics**: if the user explicitly passes `-T logprobs=false` AND configures `risk_scorer` in the YAML scorer list, fail loud with a clear error (`risk_scorer requires logprobs, but logprobs=false was explicitly set`). Don't silently override the user's input.

**User experience:**
- Calibration workflow → put `risk_scorer` in the YAML `scorer:` list. Logprobs turn on automatically. No `-T` flag needed.
- Raw logprobs workflow → `-T logprobs=true`.

No `calibration` flag. No two ways to do the same thing.

## Why this shape

- Calibration ⟹ logprobs holds, but logprobs ⟹ calibration does not. The current bundling treats them as equivalent.
- `risk_scorer` actually produces a mix of metrics: calibration (ECE, risk_calibration_error), proper scoring (Brier), and discrimination (AUC). The flag name `calibration` only describes a subset of what the scorer is for, which is part of why it misleads.
- Single source of truth: the scorer list in YAML expresses *what you're measuring*; the task figures out *what it needs to capture*. Coupling lives at the dependency boundary, not in a separate flag.
- Discoverability: `logprobs` is self-explanatory in the task signature and in `-T` strings.

## Migration

Before:

```yaml
# eval_config.yaml
calibration: true
```

After (calibration workflow — the scorer drives it):

```yaml
# eval_config.yaml
scorer:
  - name: risk_scorer
```

After (raw logprobs only, no calibration scorer):

```yaml
# eval_config.yaml
logprobs: true
```

## Files to change

Verified scope (4 files):

- `src/tools/model_organisms/inspect_task.py` — replace `calibration` parameter with `logprobs` + `top_logprobs`; update docstring (line 18 references `-T calibration=true`); implement scorer-driven auto-enable; raise on the conflict case.
- `src/tools/inspect/scorers/risk_scorer.py` — add `requires_logprobs = True` attribute on the scorer factory.
- `src/tools/inspect/setup_inspect.py:33` — replace `"calibration"` in `TASK_ARG_KEYS` with `"logprobs"` and `"top_logprobs"`.
- `tests/unit/test_model_organism_inspect_task.py` — update calls at lines 60, 65, 85 (currently `calibration=True`).

Confirmed *not* affected (no flag references found):

- Skill docs (`.claude/skills/**`) — reference "calibration metrics" / "calibration plots" downstream, never the flag.
- Scaffold templates and experiment YAMLs (`blueprints/`, `experiments/`).
- `numeric_risk_scorer` — does not need logprobs; should not declare `requires_logprobs`.

## CHANGELOG

This is a breaking change to the experiment-config surface (`-T calibration=true` no longer recognized). Add an entry to `CHANGELOG.md` as part of the PR.

## Acceptance criteria

- [ ] `tests/unit/test_model_organism_inspect_task.py` passes with the new flag names.
- [ ] An eval run with `risk_scorer` configured in YAML and no `-T logprobs` flag produces logprob entries in the eval log and computes ECE/RCE/Brier/AUC.
- [ ] An eval run with `-T logprobs=true` and no `risk_scorer` produces logprob entries in the eval log.
- [ ] An eval run with `-T logprobs=false` and `risk_scorer` configured raises a clear configuration error.
- [ ] `-T calibration=true` is rejected (unknown task arg).

## No backcompat

Per discussion, we are not preserving the `calibration` flag. Existing experiments using `-T calibration=true` will need to migrate. Repo is alpha; only test files currently use the flag.

## Out of scope (track separately)

These `risk_scorer` interpretation issues surfaced during the review and should get their own issues — they are not bugs in the rename:

- **Missing-token bias**: when the model is highly confident, the non-chosen option drops out of top-20 and the sample is silently excluded from all of `risk_scorer`'s metrics. Biases ECE/RCE toward ambiguous samples; distorts Brier and AUC by dropping the high-confidence tail.
- **`correct` vs `risk_score` divergence**: correctness comes from completion text, risk_score from logprob softmax. They can disagree under non-greedy generation.
- **Off-option mass discarded**: risk_score is conditional on the model answering with an option token, not the marginal probability of the positive class.
- **Multiclass partial**: `option_probs` is multiclass-safe; `risk_score`, `auc_score`, and `risk_calibration_error` assume binary.
- **First-token only**: multi-token answers are not scored correctly.
- **Whitespace stripping** can collide distinct tokens in some tokenizers.

msalganik:
Closes #461

# Description

The `model_organism` inspect task previously had a single `calibration: bool` flag that did two things: enabled logprob capture and auto-appended `risk_scorer`. The name described only the second behavior, leaving the underlying capability undiscoverable to anyone wanting logprobs for non-calibration analysis.

This PR replaces it with two primitives plus scorer-driven inference:

- `logprobs: bool | None = None` — capability switch. `None` means *auto*: enabled when a configured scorer declares it needs logprobs, otherwise off. `True`/`False` force.
- `top_logprobs: int = 20` — exposes the previously hardcoded value (multiclass / tokenization variants may need more).

Scorers opt in via a `requires_logprobs = True` attribute on the factory (set on `risk_scorer` only). Attribute-based detection — not name-matching — correctly skips `numeric_risk_scorer`, which reads text completion rather than logprobs.

**Conflict semantics:** `-T logprobs=false` while configuring a logprob-dependent scorer raises `ValueError` rather than silently overriding either choice.

**User experience:**
- Calibration workflow → list `risk_scorer` in YAML `scorer:`. Logprobs auto-enable.
- Raw logprobs only → `-T logprobs=true`.
- No `calibration` flag, no two ways to do the same thing.

This is a **breaking change** to the experiment-config surface: `-T calibration=true` is no longer recognized. Existing experiments must migrate. Repo is alpha; only test files used the flag.

## Files changed

- `src/tools/model_organisms/inspect_task.py` — replace `calibration` with `logprobs`/`top_logprobs`; scorer-driven auto-enable; conflict error.
- `src/tools/inspect/scorers/risk_scorer.py` — set `risk_scorer.requires_logprobs = True`.
- `src/tools/inspect/scorers/__init__.py` — add `SCORER_FACTORIES` map and `configured_scorers_require_logprobs()` helper.
- `src/tools/inspect/setup_inspect.py` — replace `calibration` in `TASK_ARG_KEYS` with `logprobs` + `top_logprobs`.
- `tests/unit/test_model_organism_inspect_task.py` — replace calibration tests with new coverage (auto-enable, explicit on/off, conflict, numeric_risk_scorer not over-included, calibration kwarg rejected).
- `CHANGELOG.md` — breaking-change entry under Unreleased.

## New Dependencies

None.

# Testing Instructions

```bash
pytest tests/unit/test_model_organism_inspect_task.py -v
pytest tests/unit -q
ruff check src/tools/model_organisms/inspect_task.py src/tools/inspect/scorers tests/unit/test_model_organism_inspect_task.py
```

## Acceptance criteria from #461

- [x] `tests/unit/test_model_organism_inspect_task.py` passes with the new flag names.
- [x] `risk_scorer` configured in YAML and no `-T logprobs` flag enables logprobs (`test_risk_scorer_auto_enables_logprobs`).
- [x] `-T logprobs=true` and no `risk_scorer` enables logprobs (`test_explicit_true_enables_logprobs`).
- [x] `-T logprobs=false` with `risk_scorer` raises a clear configuration error (`test_explicit_false_with_risk_scorer_raises`).
- [x] `calibration=...` is rejected (`test_calibration_kwarg_rejected`).

The third and fourth criteria specify behavior on a real eval run; here they are validated at the task-construction level (the surface the issue asked to fix). Logprob capture and the resulting metrics are exercised by existing scorer integration tests under `tests/unit/scorers/`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to cruijff_kit will be documented in this file.
 ### Changed
 
 - `python-markdown` added as a runtime dependency (used by the quiz renderer for intro / prompt / explanation / write-up markdown).
+- **Breaking:** `model_organism` inspect task replaces the bundled `calibration` flag with two primitives: `logprobs: bool | None` (capability switch) and `top_logprobs: int` (previously hardcoded at 20). Logprobs auto-enable when a configured scorer declares `requires_logprobs = True` (currently only `risk_scorer`). Existing experiment configs using `-T calibration=true` must migrate to placing `risk_scorer` in the YAML `scorer:` list (calibration workflow) or `-T logprobs=true` (raw logprobs only). (#461)
 
 ## [0.2.2] - 2026-04-23
 
--- a/src/tools/inspect/scorers/__init__.py
+++ b/src/tools/inspect/scorers/__init__.py
@@ -14,6 +14,15 @@ from .calibration_metrics import (
     auc_score as auc_score,
 )
 
+# Underlying scorer factories keyed by registered name. Used for capability
+# introspection (e.g., `requires_logprobs`) before instantiation.
+SCORER_FACTORIES = {
+    "match": match,
+    "includes": includes,
+    "risk_scorer": risk_scorer,
+    "numeric_risk_scorer": numeric_risk_scorer,
+}
+
 # Registry of available scorers and their constructors.
 # Task files use build_scorers() to instantiate scorers from YAML config.
 SCORER_REGISTRY = {
@@ -28,6 +37,22 @@ SCORER_REGISTRY = {
         numeric_risk_scorer(**params) if params else numeric_risk_scorer()
     ),
 }
+
+
+def configured_scorers_require_logprobs(config: dict) -> bool:
+    """Return True if any scorer in the YAML config declares it needs logprobs.
+
+    Detection is attribute-based (not name-based): a scorer factory opts in by
+    setting ``requires_logprobs = True``. This avoids over-including scorers
+    whose names happen to match a heuristic but which read text completion
+    instead of logprobs (e.g., `numeric_risk_scorer`).
+    """
+    for entry in config.get("scorer", []) or []:
+        factory = SCORER_FACTORIES.get(entry.get("name"))
+        if factory is not None and getattr(factory, "requires_logprobs", False):
+            return True
+    return False
+
 
 # Default scorers when no config is provided
 DEFAULT_SCORERS = [
--- a/src/tools/inspect/scorers/risk_scorer.py
+++ b/src/tools/inspect/scorers/risk_scorer.py
@@ -46,7 +46,7 @@ def mean_risk_score() -> Metric:
 def risk_scorer(option_tokens: list[str] = ("0", "1")):
     """
     Scorer that extracts risk scores from logprobs of the first generated token.
-    - Requires GenerateConfig(logprobs=True, top_logprobs=20) on the Task.
+    - Requires GenerateConfig(logprobs=True, top_logprobs>=2) on the Task.
     - Supports any number of option tokens (binary or multiclass).
     - For binary tasks, returns a risk_score = P(last/positive option token).
     - For all tasks, returns normalized probabilities over all option tokens.
@@ -129,3 +129,10 @@ def risk_scorer(option_tokens: list[str] = ("0", "1")):
         )
 
     return score
+
+
+# Capability marker: the unified `model_organism` task auto-enables logprobs
+# capture when a configured scorer declares this attribute. Set on the factory
+# (not the inner score function) so the task can introspect by name lookup
+# without instantiating the scorer first. See src/tools/model_organisms/inspect_task.py.
+risk_scorer.requires_logprobs = True
--- a/src/tools/inspect/setup_inspect.py
+++ b/src/tools/inspect/setup_inspect.py
@@ -30,7 +30,8 @@ TASK_ARG_KEYS = [
     "vis_label",
     "use_chat_template",
     "split",
-    "calibration",
+    "logprobs",
+    "top_logprobs",
 ]
 
 # Keys in eval_config.yaml that become --metadata args in the inspect command
--- a/src/tools/model_organisms/inspect_task.py
+++ b/src/tools/model_organisms/inspect_task.py
@@ -12,10 +12,16 @@ Usage::
         -T data_path=/path/to/dataset.json \\
         -T config_path=/path/to/eval_config.yaml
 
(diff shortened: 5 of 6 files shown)


msalganik:
@niznik-dev heads up on one judgment call: the issue specified `logprobs: bool = False`, but I went with `logprobs: bool | None = None` so the function can distinguish "user explicitly passed `-T logprobs=false`" (must raise when `risk_scorer` is configured) from "user didn't pass anything" (must auto-enable when `risk_scorer` is configured). Standard `T | None` sentinel pattern — you've seen it before. CLI surface is unchanged. Easy to revert if you'd rather express the third state differently.

niznik-dev:
This looks good to me! The CHANGELOG message is a bit verbose but I'll clean that up when I do the release tomorrow. Well done!

Issue #318, opened 2026-02-17
Add eval performance settings to scaffold-inspect


msalganik:
## Context

When running inspect-ai evaluations with HuggingFace local models, the effective batch size is controlled by `--max-connections` (default: 32). This flag was originally designed for API concurrency but is repurposed by inspect-ai's HF provider to mean "how many samples to batch into each `model.generate()` call."

Currently, `--max-connections` is not part of the scaffold-inspect workflow. There is no way to configure it in `experiment_summary.yaml`, and it does not appear in generated SLURM scripts. When we needed to increase it (e.g., to 128) for the `acs_employment_model_size_2026-02-16` experiment, we had to manually edit the SLURM scripts after scaffolding.

With 1B-3B models, the default batch size of 32 significantly underutilizes GPU memory. Increasing to 128 is safe and should improve throughput.

## What should change

1. **Add an eval performance section to `experiment_summary.yaml`** — something like `max_connections` (or a more intuitive name like `eval_batch_size`) that can be set per-experiment or per-run.

2. **Wire it through scaffold-inspect** — the generated SLURM scripts should include `--max-connections <value>` in the `inspect eval` command when specified.

3. **Set a sensible default** — 32 is conservative. Consider defaulting to 64 or 128 for local HF models, or choosing a default based on model size.

4. **Document the parameter** — explain what it controls and how to choose a value. Key points:
   - For HF local models, it controls GPU batch size (not API concurrency)
   - Larger values improve GPU utilization but increase memory usage
   - Rule of thumb: double until OOM, then back off
   - Small models (1B) on 40GB GPUs can handle 128-256; 3B models on 80GB GPUs can handle 128+

## Part of a larger effort

This is one piece of a broader set of eval performance improvements. Other potential items include:
- Automatic GPU memory-aware batch size selection
- Time limit estimation based on model size and dataset size

## Discovered during

`acs_employment_model_size_2026-02-16` — the 3B base model evaluation timed out at 30 minutes partly because the default batch size of 32 underutilized the GPU.

msalganik:
## Investigation: Why GPU utilization is low despite `--max-connections 128`

### Root cause

The inspect-ai HF provider (`inspect_ai/model/_providers/hf.py`) uses a background thread with a `queue.Queue` to batch samples. The batch fires when either:
1. It collects `max_connections` samples, OR
2. **2 seconds pass with no new sample arriving** (`queue.get(timeout=2)` at line 471)

[...]

PR #464, opened 2026-05-06
Expose --cpus_per_task knob in torchtune scaffold (#449, partial)

### LINKED ISSUE #449: Expose cpus_per_task
## For Mattie

**Problem.** PyTorch's DataLoader has a `num_workers` argument that controls how many subprocess workers prepare batches in parallel. With the default `num_workers=0`, all data prep (tokenization + chat-template formatting + collation) happens single-threaded in the main training process, blocking the GPU. The slurm scripts also default to `--cpus-per-task=1`. cruijff_kit's torchtune templates don't expose either knob, so every fine-tune we run today is effectively single-threaded data loading on a single core, even when more capacity is available.

**Why fix it.** The `consistency_curve_pilot_nth14_k20` N=1000 finetune ran at 97% CPU efficiency on a single core, while the GPU sat at 70% utilization (verifiably waiting on the CPU). Bumping just the slurm allocation does nothing if `num_workers` is unset — the workers default to 0 and the extra cores sit idle. With both knobs configurable, we expect GPU utilization to climb to 90%+ and training-only wallclock to drop ~20–25% across all our consistency-curve experiments.

**What changes after the fix.**
- `controls.cpus_per_task: 8` in `experiment_summary.yaml` reaches the generated `finetune.slurm` (`#SBATCH --cpus-per-task=8`).
- `controls.num_workers: 4` reaches both `dataset:` and `dataset_val:` blocks of the generated `finetune.yaml`.
- `controls.persistent_workers: true` does the same (saves worker-startup cost across our 3-epoch runs).
- Existing experiments that don't set these fields are unaffected: `cpus_per_task` falls back to 1 (current behavior), `num_workers` falls back to 0 (PyTorch default), `persistent_workers` falls back to false.

**Acceptance.**
1. Setting all three fields in `controls` of `experiment_summary.yaml` causes them to appear in the generated `finetune.slurm` and `finetune.yaml` for every fine-tuned run.
2. With these set, `seff <jobid>` reports `Cores: 8` and total CPU time scaled across cores; GPU utilization in `gpu_metrics.csv` reaches 90%+ during training periods.
3. Three regression tests cover the propagation path (one per knob).

**Effort estimate.** ~25 lines source across 3 files + ~30 lines tests.

---

### Investigation already done

Don't re-investigate — go straight to the patch.

- `src/tools/torchtune/templates/finetune_template.yaml` — has `dataset:` (~line 68) and `dataset_val:` (~line 84) blocks. Neither has `num_workers` or `persistent_workers`.
- The slurm template (find via `find src/tools/torchtune -name '*.slurm*'`) hardcodes `--cpus-per-task=1` and does not read from `setup_finetune.yaml`.
- `src/tools/torchtune/setup_finetune.py` — needs `--num_workers`, `--persistent_workers`, and `--cpus_per_task` CLI flags. Existing flag pattern uses argparse; place near the other dataset-related / slurm-related flags.
- `.claude/agents/scaffold-torchtune.md` — line 230 lists per-run parameter overrides. Add all three to that list and to the controls table.
- The custom recipe (`src/tools/torchtune/custom_recipes/lora_finetune_single_device_nightly.py`) constructs the dataloader from cfg — no recipe changes needed; PyTorch's DataLoader honors the YAML field directly when scaffold-torchtune writes it through.
- Concrete observation: `seff 7604383` (the pilot N=1000 run) shows CPU 36:24 of 37:25 wallclock = 97.3% efficiency on a single core, while `gpu_metrics.csv` averages 70% GPU utilization. Textbook CPU-bound symptom.

### Files to change

Source (3 files):
1. `src/tools/torchtune/templates/finetune_template.yaml` — add `num_workers` and `persistent_workers` fields to both `dataset:` and `dataset_val:` blocks.
2. `src/tools/torchtune/templates/<slurm template>.slurm` — replace hardcoded `--cpus-per-task=1` with templated value defaulting to 1.
3. `src/tools/torchtune/setup_finetune.py` — add three CLI flags (`--num_workers`, `--persistent_workers`, `--cpus_per_task`) and the corresponding template substitutions.

Docs (1 file):
4. `.claude/agents/scaffold-torchtune.md` — document all three under controls and per-run parameters.

Tests (1 file):
5. `tests/unit/test_setup_finetune_main.py` — add regression tests for the propagation path of each knob.

### Patch 1: `finetune_template.yaml`

Inside the `dataset:` block, add:
```yaml
  num_workers: {{ num_workers | default(0) }}
  persistent_workers: {{ persistent_workers | default(false) | lower }}
```
Mirror the same two lines in the `dataset_val:` block. The `|lower` filter is important — Python booleans render as `True`/`False`, but YAML wants `true`/`false`.

### Patch 2: slurm template

Find the line `#SBATCH --cpus-per-task=1` and replace with:
```
#SBATCH --cpus-per-task={{ cpus_per_task | default(1) }}
```

### Patch 3: `setup_finetune.py`

Add CLI flags near the dataset/slurm-related args:
```python
parser.add_argument("--num_workers", type=int, default=None,
                    help="DataLoader num_workers for both train and val "
                         "(default: 0, single-process data loading).")
parser.add_argument("--persistent_workers", type=str, default=None,
                    help="If 'true', keep DataLoader workers alive across "
                         "epochs. Saves spawn cost; ignored when num_workers=0.")
parser.add_argument("--cpus_per_task", type=int, default=None,
                    help="Cores requested in the generated SLURM script "
                         "(default: 1).")
```

In the args→YAML loop, add handling that writes `num_workers` and `persistent_workers` into both dataset blocks, and `cpus_per_task` into the slurm-context dict.

### Patch 4: `scaffold-torchtune.md`

Append to the per-run parameters list (line 230):
```
num_workers: {from run.parameters.num_workers, if present}
persistent_workers: {from run.parameters.persistent_workers, if present}
cpus_per_task: {from run.parameters.cpus_per_task, if present}
```

And add to the controls table around line 88-93:
```
- `controls.num_workers` - DataLoader worker subprocesses (default 0)
- `controls.persistent_workers` - Keep workers alive across epochs (default false)
- `controls.cpus_per_task` - Cores requested in the slurm script (default 1)
```

### Patch 5: `tests/unit/test_setup_finetune_main.py`

Append a new test class:
```python
class TestParallelism:
    def test_num_workers_writes_to_both_dataset_blocks(self, run_main):
        config, _ = run_main(extra_args=["--num_workers", "4"])
        assert config["dataset"]["num_workers"] == 4
        assert config["dataset_val"]["num_workers"] == 4

    def test_persistent_workers_writes_to_both_dataset_blocks(self, run_main):
        config, _ = run_main(extra_args=["--persistent_workers", "true"])
        assert config["dataset"]["persistent_workers"] in (True, "true")
        assert config["dataset_val"]["persistent_workers"] in (True, "true")

    def test_no_num_workers_means_no_field(self, run_main):
        config, _ = run_main()
        assert "num_workers" not in config["dataset"]
        assert "num_workers" not in config["dataset_val"]

    def test_cpus_per_task_writes_to_slurm(self, run_main):
        _, slurm = run_main(extra_args=["--cpus_per_task", "8"])
        assert "--cpus-per-task=8" in slurm

    def test_no_cpus_per_task_defaults_to_one(self, run_main):
        _, slurm = run_main()
        assert "--cpus-per-task=1" in slurm
```

### Test plan

Run from repo root with `cruijff` conda env active.

**a. Unit tests:**
```bash
pytest tests/unit/test_setup_finetune_main.py -v
```

**b. Integration check:**
```bash
mkdir -p /tmp/parallelism_test && cd /tmp/parallelism_test
cp /scratch/gpfs/MSALGANIK/mjs3/ck-experiments/consistency_curve_pilot_nth14_k20/Llama-3.2-3B-Instruct_N1000/setup_finetune.yaml .
# add num_workers: 4, persistent_workers: true, cpus_per_task: 8
python /home/mjs3/cruijff_kit/src/tools/torchtune/setup_finetune.py --config_file setup_finetune.yaml
grep -A4 "^dataset:" finetune.yaml          # expect: num_workers: 4 + persistent_workers: true
grep -A4 "^dataset_val:" finetune.yaml      # same
grep "cpus-per-task" finetune.slurm         # expect: --cpus-per-task=8
```

**c. End-to-end (after the next experiment runs):**
```bash
seff <jobid>  # expect Cores: 8, CPU efficiency well below 100% per core
```

**d. Lint:**
```bash
ruff check src/tools/torchtune/setup_finetune.py tests/unit/test_setup_finetune_main.py
ruff format --check src/tools/torchtune/setup_finetune.py tests/unit/test_setup_finetune_main.py
```

### PR conventions

- Branch: `<issue-number>-cpu-dataloader-parallelism`.
- PR body starts with `Closes #<issue-number>`.

### Out of scope

- Exposing `prefetch_factor` (PyTorch default 2 is fine for our short-batch case; can be added later).
- Exposing tokenizer parallelism (`HF_TOKENIZERS_PARALLELISM` env var).
- Reformatting other parts of the templates.
- The separate `batch_size_val` propagation issue is independent and should land separately.

— Claude x Matt

msalganik:
Closes #449 (partial — see comment thread on the issue for scope reduction discussion).

# Description

Adds a `--cpus_per_task` override to `setup_finetune.py` (also readable from `setup_finetune.yaml` and from `controls.cpus_per_task` in `experiment_summary.yaml`). Resolution order is **CLI > `MODEL_CONFIGS[model]["slurm"]["cpus"]` > template default 1**. The slurm template still carries `#SBATCH --cpus-per-task=1` as a placeholder; `setup_finetune.py` rewrites it to the resolved value at scaffold time.

Today every entry in `MODEL_CONFIGS` except `Llama-3.3-70B-Instruct` has `cpus: 1` — by deliberate design (commit `fbde851`, Feb 2026), since PyTorch data loading is single-threaded at `num_workers=0` and extra cores were measured to sit idle. This PR does **not** undo that decision; it adds a per-experiment override on top of it.

## What this knob is and isn't

**It is**: a per-experiment override for the cluster-citizenship vs. GPU-utilization trade-off. Allocating more cores can produce a small GPU-util bump via tokenizer threading and OpenMP threading even at `num_workers=0`. The size of that bump is workload-dependent — see "When to use" below.

**It is not**: a fix for GPU underutilization in general. The bigger lever — true parallel data loading via `num_workers > 0` — lives inside the torchtune recipe's hardcoded `StatefulDataLoader(...)` call, which can't be patched under cruijff_kit's wrapper-only design principle. That conversation is tracked in #465.

## When to use

CPU work per batch scales with sequence length and preprocessing weight. Two regimes:

- **Worth setting**: workloads with long sequences (near `max_seq_len`) or rich chat templates / heavy per-sample transforms. Tokenization is O(text length); collation and padding scale with batch × max length. These workloads spend meaningful time on the CPU side per step, so extra cores can give a small GPU-util bump even at `num_workers=0`.
- **Leave unset**: short-text workloads (capitalization, single-token outputs, simple classification). These are GPU-dominated; the CPU step is microseconds and extra cores sit idle. Use the `MODEL_CONFIGS` default (1 for almost all entries).

This per-experiment placement is correct because text-length and preprocessing weight are *experiment* properties, not *model* properties. The same model can run a short-text capitalization fine-tune and a long-context document fine-tune; the right cpu allocation differs. `MODEL_CONFIGS` correctly owns the cluster-citizenship policy ("how much idle CPU is acceptable"); the experiment correctly owns "how much CPU work my workload generates."

## Intended workflow

1. A user with a CPU-bottlenecked workload uses the override on a measurement run: `controls.cpus_per_task: 4`.
2. They capture `gpu_metrics.csv` and `seff` output, comparing to a baseline run at the `MODEL_CONFIGS` default.
3. **If the trade is good** (meaningful GPU-util gain, willing to take the per-core efficiency hit): the principled follow-up is a `MODEL_CONFIGS` default update with the measured evidence as justification, not permanent reliance on the override.
4. **If the trade is poor**: defaults stay correct, the override saw measurement use, no further change.

The override earns its keep by enabling the empirical step. Without it, the only path to test a different cpu allocation is editing `MODEL_CONFIGS` globally, which is the wrong unit of change for a measurement experiment.

## Files changed

- `src/tools/torchtune/setup_finetune.py` — new `--cpus_per_task` argparse flag (default `None`); added to `SLURM_ONLY` so it doesn't leak into `finetune.yaml`; resolution path updated at the SLURM rendering site. Help text frames this as a measurement tool, not a perf fix.
- `tests/unit/test_setup_finetune_main.py` — new `TestCpusPerTask` class with 5 propagation tests (CLI override, YAML route, fallback to `MODEL_CONFIGS`, CLI > YAML precedence, absence from `finetune.yaml`).
- `.claude/agents/scaffold-torchtune.md` — `controls.cpus_per_task` documented in both the controls list and the per-run YAML template, with the same measurement-tool framing.
- `CHANGELOG.md` — Unreleased / Added entry referencing #449 and #465.

## What this PR doesn't ship

The original issue (#449) also proposed `num_workers` and `persistent_workers` plumbing. Both require recipe-level edits (the `StatefulDataLoader` call site is hardcoded inside `_setup_data` in each of the three custom recipes, with no `cfg.get` hook). With torchtune frozen, there's no upstream path either. Under the wrapper-only design principle articulated by @msalganik in the issue thread, those changes need a deliberate exception — tracked in #465 — rather than landing inline on a perf PR.

## New Dependencies

None.

# Testing Instructions

```bash
# Unit tests (5 new + 36 existing pass)
pytest tests/unit/test_setup_finetune_main.py -v

# Lint
ruff check src/tools/torchtune/setup_finetune.py tests/unit/test_setup_finetune_main.py
ruff format --check src/tools/torchtune/setup_finetune.py tests/unit/test_setup_finetune_main.py

# Quick scaffold dry-run to confirm the rendered slurm
mkdir -p /tmp/cpus_per_task_check && cd /tmp/cpus_per_task_check
cp <some>/setup_finetune.yaml .
echo "cpus_per_task: 8" >> setup_finetune.yaml
python -m cruijff_kit.tools.torchtune.setup_finetune --config_file setup_finetune.yaml
grep cpus-per-task finetune.slurm  # expect: --cpus-per-task=8
```


niznik-dev:
Closing this in favor of #482, which covers the same `cpus_per_task`-only scope but routes the knob through `runs[].compute.cpus_per_task` rather than `controls.cpus_per_task`.

Reasoning: `cpus_per_task` is a SLURM resource — it modifies one line of the slurm script and zero lines of the torchtune config. By the existing schema invariant (`controls.*` for training hyperparameters that affect what the model learns; `runs[].compute.*` for slurm resources like `time`/`gpus`/`mem`), it sits more naturally next to its siblings than next to `epochs`/`batch_size`/`lora_dropout`. The eval-side analogue under #318 would land at `evaluation.compute.cpus_per_task`, keeping finetune and eval cleanly separate.

Adopted from this PR into #482:
- `SLURM_ONLY` addition (prevents leak into `finetune.yaml`) + matching test
- CHANGELOG entry (trimmed)
- Test reorganization into a dedicated `TestCpusPerTask` class + CLI-overrides-yaml precedence test

Deliberately not adopted:
- The longer help-string essay (kept short to match `--gpus`/`--mem` style; the "when to use" rationale lives in the CHANGELOG entry instead)
- `controls.cpus_per_task` schema slot (see above)

—MxC

Issue #474, opened 2026-05-08
Port validation-loop support to distributed LoRA recipe (lora_finetune_distributed_nightly)


niznik-dev:
## Summary

`src/tools/torchtune/custom_recipes/lora_finetune_single_device_nightly.py` implements `run_val_every_n_steps` / `dataset_val` / mid-training `Validation loss: ...` logging, but its multi-GPU counterpart `lora_finetune_distributed_stable.py` does **not**. There is no `lora_finetune_distributed_nightly.py`.

Consequence: any multi-GPU run (currently any 70B fine-tune by default) cannot get mid-training validation loss, even when the experiment design explicitly requested it.

This issue tracks porting the val loop from `_single_device_nightly` into a new `lora_finetune_distributed_nightly.py`, so the scaffold-torchtune skill can default everyone to a `_nightly` recipe regardless of GPU count.

## Context

Discovered while fixing #471 (scaffold-torchtune silently using `_stable` even when val is requested). The fix shipped for #471 is partial: it defaults single-GPU runs to `_single_device_nightly` and emits a loud scaffold-time warning when val is requested for a multi-GPU run, because `_distributed_nightly` doesn't exist. This issue closes the gap.

## Scope

Create `src/tools/torchtune/custom_recipes/lora_finetune_distributed_nightly.py` that combines:
- The FSDP/DDP plumbing from `lora_finetune_distributed_stable.py` (~640-line diff vs single_device_stable)
- The val-loop additions from `lora_finetune_single_device_nightly.py` (~150-line diff vs single_device_stable), adapted for distributed:
  - Distributed val sampler so each rank sees a disjoint slice of the val set
  - All-reduce of val loss across ranks (token-weighted: all_reduce both `sum_loss * tokens` and `tokens` separately, then divide — naive mean-of-means is wrong with uneven shards)
  - Rank-0-only logging of `Validation loss: ...`
  - `barrier()` around the val pass so straggler ranks don't deadlock
  - Verify FSDP unshard state is correct for forward-only eval

## Acceptance

- [ ] `lora_finetune_distributed_nightly.py` exists and runs to completion under `--nproc_per_node=2` on a small toy dataset
- [ ] Mid-training `Validation loss: ...` log lines appear at the configured cadence on rank 0
- [ ] Numerical correctness check: build a tiny deterministic dataset where the expected val loss is hand-computable; verify the all-reduced result matches to within fp tolerance
[...]

PR #475, opened 2026-05-08
Default scaffold-torchtune to _single_device_nightly recipe

### LINKED ISSUE #471: scaffold-torchtune silently uses _stable recipe even when run_val_every_n_steps is set (no validation runs)
## Summary

`scaffold-torchtune` accepts `run_val_every_n_steps` and `dataset_val` from `experiment_summary.yaml` and passes them through into `setup_finetune.yaml` / `finetune.yaml`. But the SLURM script it generates calls `cruijff_kit.tools.torchtune.custom_recipes.lora_finetune_single_device_stable.__main__`, which has **no validation logic at all** — those config keys are silently ignored.

The result: the user designed an experiment that was supposed to log mid-training validation loss every 400 steps; the experiment ran to completion without errors; and there is zero validation data anywhere. The mistake only becomes visible at analysis time when you go looking for the val curve and it isn't there.

## Where

- `scaffold-torchtune` (subagent) writes `tune run cruijff_kit.tools.torchtune.custom_recipes.lora_finetune_single_device_stable.__main__` into `finetune.slurm`.
- `src/tools/torchtune/custom_recipes/lora_finetune_single_device_stable.py` — does **not** reference `run_val_every_n_steps`, `dataset_val`, or anything val-related (`grep -nE 'val|VAL|validation' ...` only matches LoRA-validation utilities, not training-loop validation).
- `src/tools/torchtune/custom_recipes/lora_finetune_single_device_nightly.py:202+` — does implement `run_val_every_n_steps`, `dataset_val`, and logs `Validation loss: ...` to stdout + the metric logger.

## Reproducer

In `experiment_summary.yaml`:

```yaml
controls:
  validation_during_training: true
  run_val_every_n_steps: 400
data:
  training:
    splits: { train: 3000, validation: 5000, test: 0 }
```

Then `scaffold-experiment` → `run-experiment`. Training completes successfully, but `grep 'Validation loss' .../slurm-*.out` returns 0 hits and there are no `val_loss` entries in the wandb run. The `_stable` recipe took the config and threw the val keys away.

## Why it matters

For any experiment that relies on val-loss trajectories to assess plateaus, train/val gap, or early stopping, this silently turns the experiment into "train-only" — so:
- Plots of train/val loss overlay or train-val gap can't be generated.
- The setup `validation_during_training: true` is dishonest in `experiment_summary.yaml` (it claims something is enabled that isn't).
- For a binary task, val-only-via-inspect-ai-match scoring rounds to {0,1}, hiding any probabilistic information about how close the val output is to the right answer — exactly what mid-training val loss would have surfaced.

This bit me on `lora_rank_sweep_3B_parity_2026-05-06` — the analysis report has a whole "Important caveat — no mid-training validation loss" section explaining why goals (3) and (4) of the original experiment design can't be delivered.

## Suggested fix

Pick one of:

1. **Fail fast at scaffold time.** When `controls.run_val_every_n_steps` (or `validation_during_training: true`, or `dataset_val`) is set in `experiment_summary.yaml`, `scaffold-torchtune` should refuse to use the `_stable` recipe and either error out or auto-switch to `_nightly`. A loud warning is the minimum; a hard error is more honest.
2. **Backport val support to `_stable`.** If `_stable` is meant to be the default recipe (and `_nightly` is more experimental), the validation loop should be merged in. The `_nightly` implementation at lines 200–860 is small; it's not a deep change.
3. **Add a `recipe:` field to experiment_summary.yaml controls** and have scaffold-torchtune pick `_stable` vs `_nightly` from there explicitly, with the schema validator complaining when val keys are set against `_stable`.

Option 1 is the cheapest fix and prevents the silent-failure mode going forward. Option 2 is the right long-term fix.

## Discovered while

Analyzing the `lora_rank_sweep_3B_parity_2026-05-06` experiment in the `model_organism` project. See `analysis/report.md` in that experiment for the downstream analysis impact (whole categories of plots had to be dropped from the report).

niznik-dev:
Closes #471

# Description

Flips the `scaffold-torchtune` agent's default recipe for single-GPU runs from `lora_finetune_single_device_stable` to `lora_finetune_single_device_nightly` so that `validation_during_training: true` actually does what it says.

**Context.** The agent has been pointing every run at `_stable` since 2025-11-06. The `_stable` recipe has no validation loop, so `run_val_every_n_steps` and `dataset_val` were silently ignored at runtime. Meanwhile `make install` and `make install-dev` have force-installed torchtune nightly since 2025-11-17 — so the install path and the scaffold default have been out of sync for ~5.5 months.

**Changes.**
- `.claude/agents/scaffold-torchtune.md` — single-GPU default flipped to `_single_device_nightly`. Added a "Multi-GPU validation warning" subsection: if a multi-GPU run requests validation, the agent must emit a loud scaffold-time warning (the distributed recipe still has no val loop — see #474).
- `src/tools/torchtune/setup_finetune.py` — added a `FileNotFoundError` guard after the GPU-count auto-switch. Previously, requesting `_single_device_nightly` on a multi-GPU run would silently rewrite to `_distributed_nightly` (which doesn't exist) and fail late at SLURM runtime. Now it errors at scaffold time with a message pointing at #474.
- `tests/integration/gpu/smoke_test_torchtune.py` — flipped the hardcoded `tune run` invocation to `_single_device_nightly` so the GPU smoke test exercises the new default.
- `blueprints/capitalization/README.md` — example `setup_finetune.yaml` snippet updated.
- `docs/ARCHITECTURE.md` — recipe-list descriptions reordered so `_nightly` is presented as the default; `_distributed_stable` annotated with the #474 caveat; `_stable` reframed as an escape hatch.
- `CHANGELOG.md` — Unreleased entries.

**Out of scope (filed as #474).** Porting validation support to the distributed recipe so multi-GPU runs (currently any 70B fine-tune) can also use `run_val_every_n_steps`. Estimated ~2.5–4.5 days; needs careful token-weighted all-reduce and real multi-GPU verification. The #474 issue body and follow-up comment list every revert/update touchpoint from this PR so it's easy to clean up when distributed val lands.

## New Dependencies

None.

# Testing Instructions

- [ ] Run `make install` (or `install-dev`) on a fresh env and confirm `torchtune` resolves to a nightly build (`pip show torchtune`).
- [ ] Trigger a single-GPU workflow test (e.g. via `test the workflow` → "LoRA Comparison") and confirm:
  - generated `setup_finetune.yaml` files reference `..._single_device_nightly`
  - generated `finetune.yaml` retains `run_val_every_n_steps` and `dataset_val` (i.e. they're not silently dropped)
  - SLURM job runs to completion and `grep 'Validation loss' slurm-*.out` returns hits at the configured cadence
- [ ] Hand-test the new `setup_finetune.py` guard by pointing `--custom_recipe` at `..._single_device_nightly` and `--gpus 4`. Expected: scaffold-time `FileNotFoundError` referencing #474, not a late SLURM-time crash.
- [ ] Run the GPU smoke test (`tests/integration/gpu/smoke_test_torchtune.py`) and confirm the `_nightly` swap doesn't regress.

—MxC

### DIFF
--- a/.claude/agents/scaffold-torchtune.md
+++ b/.claude/agents/scaffold-torchtune.md
@@ -269,10 +269,18 @@ system_prompt: {from controls.system_prompt, often empty string ""}
 prompt: {from controls.prompt, e.g., "Capitalize the given word: {input}\n"}
 
 # Custom Recipe (REQUIRED)
-# Select based on model size:
-#   - Llama 1B, 3B, 8B models (1 GPU): lora_finetune_single_device_stable
-#   - Llama 70B models (multi-GPU): lora_finetune_distributed_stable
-custom_recipe: cruijff_kit.tools.torchtune.custom_recipes.lora_finetune_{single_device|distributed}_stable
+# Select based on GPU count:
+#   - Single-GPU runs (1B / 3B / 8B): lora_finetune_single_device_nightly
+#       The `_nightly` recipe supports `run_val_every_n_steps` / `dataset_val`.
+#       The `make install` and `make install-dev` targets force torchtune nightly,
+#       so this is the supported baseline for new experiments.
+#   - Multi-GPU runs (70B): lora_finetune_distributed_stable
+#       No `_distributed_nightly` exists yet (tracked in #474). If
+#       `controls.validation_during_training: true` is set for a multi-GPU run,
+#       you MUST emit a loud warning at scaffold time AND in the run log
+#       (see "Multi-GPU validation warning" below) — validation will be silently
+#       skipped because the distributed recipe has no val loop.
+custom_recipe: cruijff_kit.tools.torchtune.custom_recipes.lora_finetune_single_device_nightly  # or lora_finetune_distributed_stable for multi-GPU
 ```
 
 4. **Write file** to `{experiment_dir}/{run_directory_name}/setup_finetune.yaml`
@@ -385,6 +393,26 @@ For each run directory:
 **If model or dataset paths don't exist:**
 - Warn user but proceed (paths might be correct on compute nodes)
 - Note in log which paths couldn't be verified
+
+### Multi-GPU validation warning
+
+If a run has `gpus > 1` (i.e., uses the `_distributed_stable` recipe) AND
+`controls.validation_during_training: true` is set in `experiment_summary.yaml`,
+the distributed recipe has no validation loop and val will be silently skipped.
+
+You MUST:
+
+1. Emit a clearly visible warning to the user when reporting scaffold results, e.g.:
+   ```
+   ⚠️  WARNING: Run '{run_name}' requested validation_during_training=true but
+   uses the distributed (multi-GPU) recipe, which does not yet support
+   mid-training validation. Validation will be SKIPPED for this run.
+   Tracked in https://github.com/niznik-dev/cruijff_kit/issues/474.
+   ```
+2. Log the same warning to `logs/scaffold-torchtune.log` for that run.
+3. Do NOT silently strip `validation_during_training` from the config — leave the
+   experiment_summary.yaml as-is so the audit trail still shows what was requested.
+4. Continue scaffolding the run (treat as a non-fatal warning, not an error).
 
 ## Logging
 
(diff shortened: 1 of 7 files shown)


niznik-dev:
@msalganik Thank you for finding this one it's been an annoying mismatch for a while apparently.

PR #486, opened 2026-05-13
Wire max_connections through scaffold-inspect with default 128

### LINKED ISSUE #318: Add eval performance settings to scaffold-inspect

niznik-dev:
Closes #318

## Description

Adds optional `evaluation.max_connections` to `experiment_summary.yaml`, propagated through `scaffold-inspect` → `eval_config.yaml` → `setup_inspect.py` to render `--max-connections N` on the generated `inspect eval` command.

**Default matches inspect-ai upstream (32).** An earlier revision of this PR proposed defaulting to 128 on the theory that bigger batches would speed up HF local evals. A sensitivity test before merging showed the opposite: on Llama-3.2-3B-Instruct + ACS employment verbose (N=1000), `max_connections=256` ran 23% slower than `max_connections=8`, with 0% GPU utilization (vs 100% at mc=8) and 4–5× more VRAM consumed. Best-guess mechanism is padding waste at large batch sizes — see [investigation comment on #318](https://github.com/niznik-dev/cruijff_kit/issues/318#issuecomment-4444540248) for the full data and reasoning.

Net deliverable: configurability (no more hand-edited SLURM scripts) without changing the default, until we have evidence about where the optimum lives for cruijff_kit workloads.

## Files changed

| File | Change |
|---|---|
| `src/tools/inspect/setup_inspect.py` | `DEFAULT_MAX_CONNECTIONS = 32`, read from config with default, wire into template |
| `src/tools/inspect/templates/eval_template.slurm` | `--max-connections <MAX_CONNECTIONS>` line before `--log-dir` |
| `.claude/agents/scaffold-inspect.md` | Document optional `evaluation.max_connections`, propagation rule (only when explicitly set), updated eval_config schema |
| `.claude/skills/design-experiment/templates/experiment_summary.yaml` | Document the field with the "raising can slow evals" caveat |
| `tests/unit/test_setup_inspect.py` | 4 new tests (default, override, high value, position) |
| `CHANGELOG.md` | Unreleased entry — plumbing ships, default matches upstream, evidence linked |

## New Dependencies

None.

## Testing Instructions

- [ ] `python -m pytest tests/unit/test_setup_inspect.py` — 62 tests pass (4 new under `TestMaxConnections`)
- [ ] Render any eval SLURM via `scaffold-experiment` with no `max_connections` set → rendered script contains `--max-connections 32`
- [ ] Set `evaluation.max_connections: 64` in `experiment_summary.yaml` → propagates through scaffold-inspect to the rendered SLURM as `--max-connections 64`
- [ ] `ruff check` + `ruff format --check` on changed files — both clean

—MxC


niznik-dev:
I think it's low risk to add this but I don't think it'll get us the gains we originally expected (see issue for discussion). Filed two follow-ups for later: #487 #488

PR #495, opened 2026-05-15
Save adapter-only checkpoints, rewrite base path for offline load

### LINKED ISSUE #99: Modify inspect-ai to accept base model and adapter weights
# Describe the Feature

Currently, we can only pass a base model or merged model (base model + adapter weights) into inspect-ai. However, in order to save space and better work with what our ideal workflow would be, we should be allowed to pass the path of a base model and then adapter weights to merge into that base model without an error.

## Related Issues

#94 #98 

# Acceptance Criteria
- [ ] The Capitalization Test validation can be performed with separate folders for base model and its adapter weights

niznik-dev:
Closes #99

# Description

Replaces `stash_adapter_weights` with a save-time rewrite of `adapter_config.json`'s `base_model_name_or_path`. After each LoRA save, the recipe overwrites that field with the local absolute base-model path so the adapter dir is self-loading via transformers' native PEFT auto-detection on offline compute nodes. `save_adapter_weights_only` default flips to `True`, skipping the merged base+LoRA checkpoint and saving ~base-model-size disk per epoch.

End-to-end verified on della via `workflow_test_2026-05-15`: rank4/rank8 fine-tune outputs are **22M / 32M** each (vs 2.4G merged baseline — ~100× per checkpoint), and both evals load the rewritten adapter dir directly to **0.970 accuracy**.

### What changed

- **Custom LoRA recipes** (single-device stable + nightly, distributed stable): replace stash_adapter_files call with `rewrite_adapter_config_base_path(output_dir, epoch, base_model_path, log)`. Base path is read once in `__init__` from `cfg.checkpointer.checkpoint_dir`.
- **`custom_recipe_utils.py`**: `stash_adapter_files` replaced with `rewrite_adapter_config_base_path`.
- **`setup_finetune.py`**: `save_adapter_weights_only` default → `True`; removed `--stash_adapter_weights` arg.
- **YAML templates / refs**: `save_adapter_weights_only: True` in `finetune_template.yaml`, `yaml_refs/1B_defaults.yaml`, `yaml_refs/8B_defaults.yaml`.
- **New: `src/tools/torchtune/uncruijff_adapter.py`** — restores HF Hub repo name before exporting a checkpoint. Reads `original_repo_id.json` when present; takes `--repo-id` as a fallback (the 1B-Instruct base on della doesn't ship that marker, so the fallback isn't theoretical).
- **Docs**: `ARCHITECTURE.md` updated to describe the new flow.
- **Tests**: new `tests/unit/test_adapter_config_rewrite.py` (10 cases — rewrite, idempotence, uncruijff with/without marker, override precedence, recursion, error paths). Full suite green (1158 passed).

### Boundary note (per [[feedback_torchtune_naming_boundary]])

`adapter_config.json` is the PEFT contract, not torchtune's — we modify one field on a file PEFT defines. The base-model resolution logic stays on our side; the saved tensor weights aren't touched.

## New Dependencies

None. Relies on `transformers` ≥ 4.36 native PEFT integration, which is already in `cruijff` env.

# Testing Instructions

1. CI: `ruff check . && ruff format --check . && pytest tests/ --ignore=tests/integration/gpu` should pass.
2. Smoke test on della — full workflow:
   ```
   /design-experiment   # use workflow_test.md (LoRA Comparison)
   /scaffold-experiment
   /run-experiment
   ```
   Expect: each `{run}/artifacts/epoch_0/` ≈ 22M (rank4) or 32M (rank8) without `model.safetensors`; `adapter_config.json` should have `base_model_name_or_path` set to `/scratch/gpfs/MSALGANIK/pretrained-llms/Llama-3.2-1B-Instruct`; eval should load it without errors and produce reasonable accuracy on capitalization.
3. Export round-trip:
   ```
   python -m cruijff_kit.tools.torchtune.uncruijff_adapter <run_dir> \
     --repo-id meta-llama/Llama-3.2-1B-Instruct
   ```
   `adapter_config.json` should then have `base_model_name_or_path: meta-llama/Llama-3.2-1B-Instruct` and load correctly on any machine with HF Hub access (or the model in HF cache).

—MxC

### DIFF
--- a/.claude/agents/scaffold-torchtune.md
+++ b/.claude/agents/scaffold-torchtune.md
@@ -248,9 +248,6 @@ epochs: {from controls.epochs}
 log_every_n_steps: {use template default, typically 1}
 run_val_every_n_steps: {50 if controls.validation_during_training else 0}
 
-# Checkpoint Options
-stash_adapter_weights: 'true'  # From template default
-
 # Output configuration
 output_dir_base: {parsed from output.base_directory}
 experiment_name: {parsed from output.base_directory}
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -117,7 +117,7 @@ setup_finetune.yaml → setup_finetune.py → finetune.yaml + finetune.slurm
 
 **Custom features added to torchtune:**
 - Selective epoch saving (`epochs_to_save`)
-- Adapter-only saves with self-loading offline (rewrite `adapter_config.json` base path after save; `uncruijff_adapter` restores HF Hub name for export)
+- Adapter-only saves with self-loading offline (rewrite `adapter_config.json` base path after save; `port_cruijff_adapter` restores HF Hub name for export)
 - Custom metrics integration via `src/utils/finetune_custom_metrics.py`
 - Validation during training (requires nightly build)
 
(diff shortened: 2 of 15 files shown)


niznik-dev:
@sarahepedersen @msalganik This is a somewhat significant change that you should read about before pulling main again. The tl;dr is that we can get away with saving just adapter weights if we change a config to point to the path where our pretrained-llms live instead of specifying the HF name of the model. The catch is if we need to move the model somewhere else to share it off of della, there's a short script to run in order to reverse what we did to make life easier. Also save_adapter_weights_only is back to True by default. Let me know if you have questions.

PR #496, opened 2026-05-18
Thread evaluation.temperature and max_tokens; reject temperature=0

### LINKED ISSUE #470: Audit experiment_summary.yaml → scaffolding field coverage
## Summary

`experiment_summary.yaml` lets users specify `evaluation.temperature` (and the design-experiment skill writes it), but `setup_inspect.py` does not thread that value into the inspect-ai task. The task therefore falls back to its hard-coded default (`1e-7`), and the user's stated temperature is silently ignored.

## Where

- `src/tools/inspect/setup_inspect.py:27` — `TASK_ARG_KEYS` does not include `"temperature"`:

  ```python
  TASK_ARG_KEYS = [
      "data_path",
      "config_path",
      "vis_label",
      "use_chat_template",
      "split",
      "calibration",
  ]
  ```

- `src/tools/model_organisms/inspect_task.py:35` — task default is `temperature: float = 1e-7`.

Result: anything written under `evaluation.temperature` (e.g. `0.0`) never appears as `-T temperature=...` on the inspect command line. There is no warning at scaffold time.

## Reproducer

1. In `experiment_summary.yaml`, set:
   ```yaml
   evaluation:
     temperature: 0.0
   ```
2. Run `scaffold-experiment`.
3. Inspect any generated `eval/{task}_epoch{N}.slurm` — there is no `-T temperature=0.0` argument, and the task uses `1e-7`.

## Why it matters

For tasks where the model output is a single token compared by `match` (e.g. `model_organism` parity / bits), `1e-7` vs `0.0` is operationally equivalent, so the divergence is invisible. But for any task where temperature affects scoring (longer generations, sampling-based scorers, calibration analyses), the user's intent is silently overridden — the eval log will report a temperature the user never chose, and results across experiments designed with different temperatures will not actually differ at runtime.

## Suggested fix

Two minimal changes:

1. Add `"temperature"` to `TASK_ARG_KEYS` in `src/tools/inspect/setup_inspect.py` so it is rendered as `-T temperature=<value>` whenever present in `eval_config.yaml`.
2. Have `scaffold-inspect` write `evaluation.temperature` into each run's `eval/eval_config.yaml` (it currently does not propagate this field).

Open question: should `temperature` always be required in `eval_config.yaml`, or remain optional and fall back to the task default? I'd lean optional-with-warning at scaffold time so existing configs keep working but the silent-drop behaviour goes away.

## Discovered while

Scaffolding the `lora_rank_sweep_3B_parity_2026-05-06` experiment (`model_organism` project). The `scaffold-inspect` subagent flagged it as a non-blocking advisory; filing here so it doesn't get lost.

niznik-dev:
Closes #470

# Description

`experiment_summary.yaml` declares `evaluation.temperature` as REQUIRED, but the value never reached the inspect-ai task — `temperature` was missing from `setup_inspect.py`'s `TASK_ARG_KEYS`, the scaffold-inspect agent didn't write it into `eval_config.yaml`, and no alarm fired anywhere. The task's hard-coded `1e-7` default silently won. For `match`-scored tasks (parity, bits, capitalization) the divergence was invisible; for anything sampling-based or calibration-related (Matt's risk_scorer / ACS work) it silently overrode user intent. The audit found `max_tokens` had the same plumbing gap.

The "Honest defaults" milestone calls this out by name — silent override of user-stated configuration is exactly the bug class we're closing.

## Audit table

| Field | In schema | In TASK_ARG_KEYS | Reached task? | Fixed here |
|---|---|---|---|---|
| `temperature` | REQUIRED | ❌ | ❌ → task default `1e-7` | ✅ |
| `max_tokens` | (absent — added) | ❌ | ❌ → task default | ✅ |
| `data_path`, `vis_label`, `use_chat_template`, `split`, `top_logprobs` | varies | ✅ | ✅ | (already worked) |
| `seed` | absent | ❌ | N/A — not a `@task` param anywhere | follow-up |

## Changes

**`src/tools/inspect/setup_inspect.py`**
- Add `temperature` and `max_tokens` to `TASK_ARG_KEYS` — they now render as `-T temperature="..." \` lines in the SLURM script via the existing `build_task_args` path. No new code path.
- Add `KNOWN_STRUCTURAL_KEYS` set + unknown-key `warnings.warn` in `load_eval_config`. Catches future drift between scaffold-inspect's writer and the downstream consumers (e.g. a typo'd field or a field added to eval_config.yaml but forgotten in `TASK_ARG_KEYS`).
- Hard-error in `load_eval_config` when `temperature <= 0`. HF generators in inspect-ai's pipeline can fail or misbehave at exactly 0 — the `@task` defaults use `1e-7` for this reason. Fail fast at scaffold time, not at SLURM runtime hours later.

**`.claude/agents/scaffold-inspect.md`**
- Eval_config.yaml schema's "Optional keys" block now shows `temperature` and `max_tokens`.
- "Populating eval_config.yaml" gains bullets telling the agent to copy `evaluation.temperature` (always) and `evaluation.max_tokens` (if set) from `experiment_summary.yaml`.

**`.claude/skills/design-experiment/templates/experiment_summary.yaml`**
- Fix actively misleading `temperature: float # REQUIRED: ... (typically 0.0)` comment. New comment recommends `1e-7` and explains why not exactly 0 — `setup_inspect.py` now rejects 0.
- Add `max_tokens` as OPTIONAL with documented per-task defaults.

## What this does NOT do (deferred)

- **`seed` propagation** — no `@task` function currently accepts a `seed` parameter, so this would require editing every task body in `blueprints/` and `src/tools/model_organisms/`. Separate scope.
- **Fully-deterministic propagation** — the experiment_summary.yaml → eval_config.yaml hop still goes through scaffold-inspect's prose. The unknown-key warning catches drift but not absence (a field present in experiment_summary.yaml that the agent forgot to write). The structural fix would be having `setup_inspect.py` read experiment_summary.yaml directly for flat-copy fields, the way `setup_finetune.py` reads the equivalents. Bigger refactor; worth a follow-up issue alongside the broader "move judgment into code" trajectory.

## New Dependencies

None.

# Testing Instructions

Verified locally in three scenarios (all pass on this branch):

1. **End-to-end render** — feed a minimal `eval_config.yaml` with `temperature: 1.0e-7` and `max_tokens: 5` through `setup_inspect.py`. Rendered SLURM contains:
   ```
   -T temperature="1e-07" \
   -T max_tokens="5" \
   ```

2. **Temperature-zero check** — `eval_config.yaml` with `temperature: 0.0` raises:
   ```
   ValueError: eval_config.yaml has temperature=0.0. Must be strictly positive
   (use 1e-7 for near-greedy decoding; exactly 0 can fail in HF generators).
   ```
   Before any SLURM script is written.

3. **Unknown-key warning** — `eval_config.yaml` with a stray field (e.g. `made_up_field: 42`) emits:
   ```
   UserWarning: eval_config.yaml has keys not consumed by setup_inspect.py
   or the task: ['made_up_field']. They will not reach the eval. ...
   ```

4. **Lint** — `ruff check` and `ruff format --check` both pass on `setup_inspect.py`.

5. **Back-compat** — older `eval_config.yaml` files without `temperature` / `max_tokens` still render correctly (omitted → no `-T` line, task default applies). No warning for missing fields; only for unknown ones.

For a full integration check, the cruijff_kit workflow tests (`test the workflow` → LoRA Comparison variant) will exercise the design → scaffold → run path end-to-end. I haven't run that here since it submits real SLURM jobs; happy to kick one off if you want it before merge.

—MxC

### DIFF
--- a/.claude/agents/scaffold-inspect.md
+++ b/.claude/agents/scaffold-inspect.md
@@ -263,6 +263,8 @@ output_dir: /outputs/run1/artifacts/
 data_path: /data/acs_income.json
 vis_label: 1B_ft
 use_chat_template: "true"
+temperature: 1.0e-7    # propagated from evaluation.temperature (must be > 0)
+max_tokens: 5          # propagated from evaluation.max_tokens (if set)
 
 # Metadata (--metadata key=value)
 epoch: 0
@@ -306,6 +308,8 @@ Extract values from setup_finetune.yaml (fine-tuned runs) or experiment_summary.
   Log the resolved path into `logs/scaffold-inspect.log` so the audit trail captures exactly which file backed this evaluation.
 - `system_prompt`: From the run's configuration (may vary per run!)
 - `scorer`: From `evaluation.scorer` in experiment_summary.yaml
+- `temperature`: From `evaluation.temperature` in experiment_summary.yaml (REQUIRED in the schema). Always propagate. Must be > 0; `setup_inspect.py` will hard-error on `<= 0` because HF generators can fail at exactly 0.
+- `max_tokens`: From `evaluation.max_tokens` in experiment_summary.yaml (OPTIONAL). Propagate only if explicitly set; otherwise omit and the @task function's per-task default applies.
 
 #### setup_inspect.py Usage
 
--- a/.claude/agents/scaffold-inspect.md
+++ b/.claude/agents/scaffold-inspect.md
@@ -308,7 +308,7 @@ Extract values from setup_finetune.yaml (fine-tuned runs) or experiment_summary.
   Log the resolved path into `logs/scaffold-inspect.log` so the audit trail captures exactly which file backed this evaluation.
 - `system_prompt`: From the run's configuration (may vary per run!)
 - `scorer`: From `evaluation.scorer` in experiment_summary.yaml
-- `temperature`: From `evaluation.temperature` in experiment_summary.yaml (REQUIRED in the schema). Always propagate. Must be > 0; `setup_inspect.py` will hard-error on `<= 0` because HF generators can fail at exactly 0.
+- `temperature`: From `evaluation.temperature` in experiment_summary.yaml (REQUIRED in the schema). Always propagate. Must be strictly > 0; `setup_inspect.py` hard-errors on `<= 0` because inspect-ai's HF provider runs with `do_sample=True`, which makes HuggingFace's `TemperatureLogitsWarper` do `scores / temperature` and raise `ValueError` at zero (division by zero → inf/NaN logits).
 - `max_tokens`: From `evaluation.max_tokens` in experiment_summary.yaml (OPTIONAL). Propagate only if explicitly set; otherwise omit and the @task function's per-task default applies.
 
 #### setup_inspect.py Usage
--- a/.claude/agents/scaffold-inspect.md
+++ b/.claude/agents/scaffold-inspect.md
@@ -308,7 +308,7 @@ Extract values from setup_finetune.yaml (fine-tuned runs) or experiment_summary.
   Log the resolved path into `logs/scaffold-inspect.log` so the audit trail captures exactly which file backed this evaluation.
 - `system_prompt`: From the run's configuration (may vary per run!)
 - `scorer`: From `evaluation.scorer` in experiment_summary.yaml
-- `temperature`: From `evaluation.temperature` in experiment_summary.yaml (REQUIRED in the schema). Always propagate. Must be strictly > 0; `setup_inspect.py` hard-errors on `<= 0` because inspect-ai's HF provider runs with `do_sample=True`, which makes HuggingFace's `TemperatureLogitsWarper` do `scores / temperature` and raise `ValueError` at zero (division by zero → inf/NaN logits).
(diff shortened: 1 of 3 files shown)


niznik-dev:
Follow-up commit `b2faaba` replaces the vibes-based "HF generators can fail or misbehave" wording with the actual mechanism. Briefly:

inspect-ai's HF provider defaults to `do_sample=True`. With `do_sample=True`, `model.generate(temperature=X, ...)` builds a `transformers.generation.logits_process.TemperatureLogitsWarper`, whose `__init__` hard-rejects `temperature <= 0`:

```python
if not isinstance(temperature, float) or not (temperature > 0):
    raise ValueError(
        f"`temperature` (={temperature}) has to be a strictly positive float, "
        "otherwise your next token scores will be invalid."
    )
```

(Reason: the warper's `__call__` does `scores / self.temperature` — a zero divisor produces inf/NaN logits and a broken softmax.)

The scaffold-time check in this PR catches that before any SLURM job is submitted. The new error message names `TemperatureLogitsWarper` and offers both fixes — `1e-7` for near-greedy, or `-M do_sample=false` for true greedy (temperature ignored, sampling disabled).

—MxC

Issue #437, opened 2026-04-22
Folder reorganization (atomic): projects/, src/, and model_organism split


niznik-dev:
## Summary

Committed plan for a folder reorganization aligning cruijff_kit with a unified "blueprints in repo, runs on scratch" model. **Atomic PR** — one big change, internally chunked for reviewability.

_This issue body supersedes the earlier draft. Full plan doc is in `edi-ccore/reviews/folder_reorg_plan_2026-04-22.md` (private); the relevant content is summarized below for group visibility._

### Top-level repo layout (post-reorg)

```
cruijff_kit/
├── pyproject.toml               # src layout
├── src/cruijff_kit/             # all library code
│   ├── tools/
│   │   ├── model_organisms/     # was sanity_checks/model_organisms/
│   │   ├── inspect/
│   │   ├── torchtune/
│   │   ├── slurm/
│   │   └── experiment/
│   ├── utils/
│   └── text_gen/
├── projects/                    # blueprints (read-only, never executed from)
│   ├── capitalization/
│   ├── folktexts/
│   └── model_organism/
├── tests/
├── docs/
└── [README, CLAUDE.md, etc.]
```

### Blueprint canonical shape (5 slots, 3 required)

```
projects/{name}/
├── README.md          # REQUIRED
├── inspect_task.py    # REQUIRED
├── generate_data.py   # REQUIRED
├── modifiers/         # optional (project-specific data transforms)
└── baseline.py        # optional (non-LLM comparison)
```

Earlier drafts had `lib/` and `templates/` slots; both dropped as YAGNI (no current project needs them; when primitives stabilize they graduate to `src/cruijff_kit/tools/`).

### Scratch-side structure (Option 3b — per-experiment snapshot)

```
ck-projects/{project}/
├── README.md                         # user's cross-run notes
├── findings_across_runs.md           # optional user docs
├── 2026-04-25_parity_bits_r4/
│   ├── code/                         # blueprint snapshot at run time
│   ├── checkpoints/
│   ├── logs/
│   ├── eval_output/
│   └── plots/
└── ...
```

**No living `code/` at project root** — each experiment owns its own snapshot. Only two sources of truth: the repo blueprint OR a specific experiment's snapshot. Scientific reproducibility without path archaeology.

`scaffold-experiment` skill performs the blueprint copy-out, aligned with the data-generation step Sarah added in PR #414/#423 chunks 5–7.

## Naming decisions

[...]

PR #508, opened 2026-05-19
Merge Emnet's continuous branch

### LINKED ISSUE #366: Add continuous variable prediction to inspect evaluation
## Summary

Extend the inspect-ai evaluation infrastructure to support continuous (regression) variable prediction, alongside the existing binary classification support.

## Motivation

All current inspect tasks and scorers assume binary classification (two discrete options, logprob extraction, AUC/Brier/ECE metrics). Supporting continuous variable prediction opens up new experiment types — e.g., predicting income values, ages, or other numeric outcomes rather than discretizing them into bins.

## Current State

The evaluation stack is binary-classification-only:

| Component | Location | Binary Assumption |
|-----------|----------|-------------------|
| `risk_scorer()` | `tools/inspect/scorers/risk_scorer.py` | Extracts P(token) for two option tokens |
| `numeric_risk_scorer()` | `tools/inspect/scorers/risk_scorer.py` | Parses float output but thresholds at 0.5 |
| Calibration metrics | `tools/inspect/scorers/calibration_metrics.py` | ECE, Brier, AUC — all binary |
| `detect_metrics()` | `tools/inspect/viz_helpers.py` | Classifies as accuracy vs supplementary (binary) |
| Report generator | `tools/inspect/report_generator.py` | Confusion matrix, balanced accuracy, F1 |
| `summary_binary.py` | `tools/inspect/` | Name says it all |

### New Components
- [ ] **Regression scorer** — parse arbitrary numeric output from model, compute per-sample error, store predicted/actual values in metadata
- [ ] **Regression metrics** — MAE, MSE/RMSE, R², Pearson/Spearman correlation
- [ ] Register new scorer(s) in `SCORER_REGISTRY` (`tools/inspect/scorers/__init__.py`)

### Updated Components
- [ ] **`detect_metrics()`** — distinguish regression vs classification task types
- [ ] **Report generator** — regression-appropriate summary stats (no confusion matrix; instead show error distributions, correlation)
- [ ] **Visualization helpers** — handle continuous predictions (scatter plots, residual plots vs heatmaps/confusion matrices)
- [ ] **Analysis pipeline** — `analyze-experiment` skill should detect task type and route to appropriate analysis

### Design Decisions to Make
- How does the task declare itself as regression vs classification? (metadata field? scorer type inference?)
- What output format do we expect from the model for continuous predictions? (free-form float? structured?)
- Do we need calibration analogues for regression? (prediction intervals, reliability diagrams for continuous outcomes)

## Assignee

Emnet is leading this work and will push a branch soon.

## Related

- `tools/inspect/scorers/` — existing scorer infrastructure
- `tools/inspect/report_generator.py` — report generation
- `tools/inspect/viz_helpers.py` — metric detection and visualization

sarahepedersen:
# Description

thank you @EmnetSy !
- Adds continuous scorer to inspect tools to parse numeric output from a given number of output tokens 
- Pipes through the ACS tasks for continuous variables (income, travel time, age, etc.) 
- Computes summary metrics (mse, R-squared) to evaluate continuous predictions 

# Testing Instructions

- test file `tests/unit/scorers/test_continuous_scorer.py`
- ran sample experiment with base models and fine-tuned models and was able to get output from both


niznik-dev:
**`inspect_task_acs_continuous.py:26`** — `from cruijff_kit.tools.inspect import hf_prefill_patch` will `ImportError`.

The `hf_prefill_patch` reference predates the big refactor (#437) and nothing in the current tree references it anymore. Safe to just delete the line.

—MxC

niznik-dev:
Smoke-tested the continuous data path: convert.py with no `--target-threshold` writes `output` as a numeric string (e.g. `"62096.0"`), and the scorer parses it cleanly (25/25 on a synthetic test set).

One mismatch to flag, though: `README_continuous.md` advertises `WKHP` (hours worked) and `JWMNP` (commute time) as available targets, but `src/tabular_to_text_gen/schemas/acs_example.yaml` only defines `AGEP`, `COW`, `SCHL`, `MAR`, `OCCP`, `SEX`, `PINCP`, `ST`. So `acs_age` and `acs_income_continuous` work today; `acs_hours` and `acs_commute` would fail at convert time until the schema gets those columns added.

Either add `WKHP` and `JWMNP` entries to the schema, or trim the README's target table to what actually works.

—MxC

niznik-dev:
## Notes / future directions

A few open-ended observations from a final pass — none of these need to land in this PR, but worth jotting down so we don't lose track once continuous evals get more use.

**1. Score value on parse failure.** The scorer returns `value=INCORRECT` (a binary-classification idiom) when parsing fails, then `value=abs(error)` on success. Those land in the same default-score field, so the aggregate gets a little weird. Switching the failure case to `float("nan")` (or overriding the default aggregator) would make the eval-log default score field cleaner. Nothing breaks today — `mae`/`rmse`/`r_squared` are the metrics that matter.

**2. `parse_rate` granularity.** Currently it conflates "model output didn't parse" with "target didn't parse." In practice `convert.py` shouldn't emit unparseable targets, but if it ever does (hand-written datasets, missing data), we'd want to know which side failed. Splitting into `prediction_parsed` / `target_parsed` in metadata would make this debuggable.

**3. Trailing `.0` in labels.** `compute_label` returns `str(target_value)`, which gives integer-valued floats a `.0` suffix (e.g. `"62096.0"`). The regex eats it cleanly, so this is purely cosmetic, but the model will learn to emit `.0` after every prediction. Could be worth stripping in `compute_label` for tidiness.

**4. Test loop hygiene.** The tests use `asyncio.new_event_loop().run_until_complete(...)` per call, which spins up a fresh loop each time. `pytest-asyncio` would be lighter weight if/when the test count grows.

Also worth a glance: #509 captures the downstream summarize/analyze story, which is the bigger follow-up.

Great work overall — the scorer's metric guards (NaN on empty / constant-targets / <2 samples) are the kind of detail that pays off later.

—MxC

niznik-dev (dismissed):
Approving this pending the schema updates we discussed on Slack. Ping me if your changes require me to re-approve but I'll let you merge when you're satisfied.

niznik-dev:
@sarahepedersen Can we get this one in today? Happy to help if you don't have the bandwidth

sarahepedersen:
> @sarahepedersen Can we get this one in today? Happy to help if you don't have the bandwidth

yep - I will change the schema and do a final pass on it! I wanted to triple check that everything's working as expected with the tokenizer & look over some of the Inspect individual predictions. aiming to have it in by 3ish

Issue #54, opened 2025-06-20
Integrate embeddings computation into metrics logging


drigobon:
# Describe the Feature

Add functionality to our fine-tuning training loop to save hidden state embeddings of input data. 

## Related Issues

Current metrics are logged in `_loss_step`, unclear if this is being run during validation or training steps.

# Acceptance Criteria
- [ ] Saves (csv? json?) embeddings files for comparison and analysis later.
- [ ] Expand config to include option of not saving embeddings matrices
- [ ] Ensure true labels / targets are properly associated with corresponding embeddings 
- [ ] Differentiate between training and validation embeddings?

PR #516, opened 2026-05-21
Close #465: adopt recipe patching policy and guard in-train() patches

### LINKED ISSUE #465: Reconcile the wrapper-only design principle with current torchtune recipe patches
## Context

In conversation around #449 (PR #464), @msalganik articulated a design principle for cruijff_kit:

> All fine-tuning code is written and tested by torchtune. cruijff_kit's role is to build wrappers around that, not to take ownership of training-loop or training-adjacent code.

This is a strong correctness boundary: torchtune has tested its dataloader, distributed training, gradient accumulation, etc. If cruijff_kit patches the recipe, we now own the verification of the patched paths.

## Current state

The principle is the *target* but not the *current* state. There are **10 marked `# !--- cruijff_kit patch ---!` blocks per recipe**, in each of the three custom recipes:

- `lora_finetune_single_device_nightly.py`
- `lora_finetune_single_device_stable.py`
- `lora_finetune_distributed_stable.py`

= 30 patch blocks total. Examples include `stash_adapter_files` (movable — pure file shuffling after `save_checkpoint`), `epochs_to_save` filtering (inside the training loop), config reads, validation hooks. Some are movable to a post-job step or wrapper layer; some genuinely need to run inside the training loop and read training state.

With torchtune frozen (security patches only), these forks aren't a transitional state — they're our permanent vendored copy.

## Why this matters now

Issue #449 wanted to expose `num_workers` and `persistent_workers` as scaffold knobs to fix GPU underutilization. The DataLoader is constructed inside `_setup_data` with no hook. Under the principle, that change isn't shippable; without it, the issue can't fully close. PR #464 ships only the principle-clean piece (`cpus_per_task`); the rest is upstream-blocked-and-deferred to this conversation.

This pattern is going to recur. Every future feature that needs to reach into recipe code will hit the same fork. So we should decide deliberately, once:

## Three options for how to apply the principle going forward

1. **Strict.** Don't add new patches. Open follow-ups to extract every existing patch that's movable (probably ~half). Document the irreducible remainder as "we accept this divergence because no wrapper-layer mechanism exists." Performance ceilings that require recipe access (like #449's `num_workers`) become accepted limitations.

2. **Aspirational + case-by-case exceptions.** Each new patch needs an explicit "is this movable?" check. If movable → move it. If not, and the value is high → add it with a marked block, and the marked block is the audit trail. Existing patches stay unless someone has bandwidth to extract them.

3. **Status quo.** Don't add new patches *or* extract existing ones. Defer perf work that needs recipe access indefinitely.

## What I'd find useful from this discussion

- Which option (1/2/3) you want, and what triggers an exception under (2) if that's the choice.
- Whether the existing `stash_adapter_files` patch is (a) regret, (b) grandfathered, (c) within the principle as you read it. The answer changes the bar for similar future work.
- Whether to file a separate tracking issue for "extract `stash_adapter_files` from recipes" as a smoke-test for the strict path.
- Whether the principle should be written into `CLAUDE.md` so future Claudes (and humans) don't have to rediscover it case-by-case.

## Out of scope here

- Specific implementation of `num_workers` (covered by deferred work in #449).
- Changes to torchtune itself (frozen).
- General fork-management tooling.

cc @niznik-dev

niznik-dev:
Closes #465

## Summary

Adopts the recipe patching policy from `edi-ccore/docs/465_recipe_patch_rule_plan.md` and applies it to the three HIGH-risk in-`train()` patches identified during the audit:

- **P1 (embeddings hook, nightly only)**: removed. The feature is broken (`.item()` on multi-element tensors crashes immediately), incomplete (no save to disk), runs the forward pass twice per batch, and leaves the model in `eval()` mode after the first call. Also unused in any current config. A follow-up issue will track a corrected reimplementation. Original feature came from #54 / #77.
- **P2 (`calculate_custom_metrics`, all 3 recipes)**: wrapped in `try/except` with auto-disable on failure, so a buggy metric can't kill an entire training run.
- **P3 (`epochs_to_save`, all 3 recipes)**: new `validate_epochs_to_save()` helper that fails loud at init on out-of-range indices, empty lists, and wrong types. Previously these silently produced zero-checkpoint runs that only emitted "Skipping checkpoint save" logs every epoch.

## New Dependencies

None.

## Testing Instructions

- \`python -m pytest tests/unit/test_epochs_to_save.py\` — 11 new cases, all pass
- \`python -m pytest tests/unit/test_adapter_config_rewrite.py\` — adjacent regression check on \`custom_recipe_utils\`, still passes
- \`ruff check\` and \`ruff format --check\` are clean on all touched files

—MxC


niznik-dev:
Filed #517 to track the embeddings extraction reimplementation. —MxC

PR #523, opened 2026-05-28
Strip in-train() patches from custom recipes (#465 wrapper-only)

niznik-dev:
Relates to #465 (does not close — policy thread stays open).

## Summary

Implements the wrapper-only direction agreed on the #465 thread: training math inside the recipe boundary is now identical to upstream torchtune for all three custom recipes. The remaining cruijff_kit responsibility — post-save adapter housekeeping — moves to a new wrapper-layer step at `src/tools/torchtune/post_finetune.py`, invoked from `finetune.slurm` after \`tune run\` succeeds.

Per Matt's call-out: _\"a new pull request will address these and test how it goes.\"_ This is the test-it PR.

## Features removed from the recipes

| F# | Feature | Disposition |
|---|---|---|
| F1 | \`epochs_to_save\` / \`save_last_epoch_only\` / \`validate_epochs_to_save\` | Removed. Upstream torchtune saves every epoch; combine with \`save_adapter_weights_only: True\` (default) to keep disk usage modest. |
| F2 | \`rewrite_adapter_config_base_path\` / \`stash_adapter_files\` | **Moved** to \`src/tools/torchtune/post_finetune.py\`. SLURM script now runs \`python -m cruijff_kit.tools.torchtune.post_finetune --config finetune.yaml\` after training. |
| F3 | Custom-metrics framework (recipe-side + \`src/utils/finetune_custom_metrics.py\`) | Removed. No live consumers in this repo. |
| F4 | \`num_workers\` / \`persistent_workers\` (recipes, yamls, CLI flags) | Removed. Measured throughput gains were minimal. |
| F5 | \`tqdm_miniters\` | Removed. Defaults are sufficient. |
| F6 | Dead \`setup_logger\` import block | Removed. The \`setup_logger\` module itself stays; only the recipe binding was dead. |

## Files touched

- 3 custom recipes (all patch markers gone; \`grep \"cruijff_kit patch\" src/tools/torchtune/custom_recipes/\` returns nothing)
- 3 yamls (\`1B_defaults\`, \`8B_defaults\`, \`finetune_template\`)
- \`setup_finetune.py\` — argparse + config-merge entries for removed flags
- \`custom_recipe_utils.py\` — drop \`validate_epochs_to_save\`; F2 helpers remain (now imported by \`post_finetune.py\`)
- \`finetune_template.slurm\` — adds the post-finetune step gated on \`TUNE_EXIT_CODE\`
[...]

Issue #529, opened 2026-05-28
ζ (F6) — Remove dead setup_logger import block from recipes


niznik-dev:
Part of #465 (wrapper-only principle). Carved out of closed PR #523.

## Context

Each custom recipe has a 4-line block at the top that imports `setup_logger` from `cruijff_kit.utils.logger` and binds a module-level `logger`, but **the resulting `logger` variable is never used anywhere in the recipe**. It's pure dead code. Removing it makes the recipes that much closer to upstream-identical at the top of file.

The `src/utils/logger.py` module itself stays — it's still imported by two other modules in the inspect subtree.

## Scope

- **Recipes affected**: `lora_finetune_single_device_stable`, `lora_finetune_single_device_nightly`, `lora_finetune_distributed_stable` (~4 lines each)
- The dead block looks roughly like:
  \`\`\`python
  # Feature: custom_logger - Custom logging setup
  from cruijff_kit.utils.logger import setup_logger

  # Set up logging
  logger = setup_logger(__name__)
  \`\`\`
- **No YAML changes** (no config flag)
- **No \`setup_finetune.py\` changes** (no CLI flag)
- **No tests affected**

## Reference implementation

[PR #523](https://github.com/niznik-dev/cruijff_kit/pull/523) (closed) contains the working diff. The branch `465-strict-wrapper-principle` can be cherry-picked from.

## Acceptance criteria

- [ ] No \`setup_logger\` import in any of the three custom recipes
- [ ] \`src/utils/logger.py\` is unchanged (still importable, still used by the inspect-tree consumers)
- [ ] \`pytest tests/unit/\` passes
- [ ] No regression on a smoke run — but no SLURM run required to validate (\`pytest\` is sufficient)

—MxC

Issue #525, opened 2026-05-28
β (F1) — Remove epochs_to_save / save_last_epoch_only / validate_epochs_to_save from recipes


niznik-dev:
Part of #465 (wrapper-only principle). Carved out of closed PR #523.

## Context

The custom recipes patch `train()` and `__init__()` to support **selective epoch saving** via the cruijff_kit-specific config keys `epochs_to_save` (list of epoch indices) and `save_last_epoch_only` (bool), with a `validate_epochs_to_save()` helper that fails loud on bad input. The #465 audit found that this is exactly the silent-corruption class we want to eliminate — a misformatted `epochs_to_save` could (and did) yield zero-checkpoint runs.

Per the wrapper-only policy, training-loop state should match upstream torchtune. The fix: drop the selective-save machinery entirely. Upstream torchtune saves every epoch; users who care about disk should set `save_adapter_weights_only: True` (already the default).

## Scope

- **Recipes affected**: `lora_finetune_single_device_stable`, `lora_finetune_single_device_nightly`, `lora_finetune_distributed_stable`
- **`custom_recipe_utils.py`**: drop `validate_epochs_to_save()`
- **YAML refs affected**: `src/tools/torchtune/yaml_refs/1B_defaults.yaml`, `8B_defaults.yaml`, `templates/finetune_template.yaml`
- **`setup_finetune.py`**: drop argparse + config-merge entries for `epochs_to_save` / `save_last_epoch_only`
- **Tests affected**: delete `tests/unit/test_epochs_to_save.py`; drop the `test_parse_epochs_*`, `test_suggest_checkpoint_subset_*`, and checkpoint-warning tests from `tests/unit/test_setup_finetune.py`
- **User-facing config flags removed**: `epochs_to_save`, `save_last_epoch_only`
- **Docs**: note in `CHANGELOG.md` + `docs/ARCHITECTURE.md` that selective saving is gone; recommend pairing every-epoch saves with `save_adapter_weights_only: True`

## Reference implementation

[PR #523](https://github.com/niznik-dev/cruijff_kit/pull/523) (closed) contains the working diff. The branch `465-strict-wrapper-principle` can be cherry-picked from.

**Note**: this issue depends on α (F2) landing first — once F2 has moved adapter housekeeping out of the recipe, the F1 removal becomes a pure delete.

## Acceptance criteria

[...]

PR #530, opened 2026-05-28
α (F2) — Move adapter housekeeping to post_finetune wrapper

### LINKED ISSUE #524: α (F2) — Move adapter housekeeping (rewrite_adapter_config_base_path, stash_adapter_files) to wrapper layer
Part of #465 (wrapper-only principle). Carved out of closed PR #523.

## Context

Today the custom torchtune recipes patch `save_checkpoint()` (or equivalent) to (a) rewrite `adapter_config.json`'s `base_model_name_or_path` so downstream inspect-ai evals can load the adapter from a portable path, and (b) stash a few adapter-companion files into each epoch directory. Per the #465 audit and the wrapper-only principle, post-`save_checkpoint` file operations should not live inside the recipe — they belong in a wrapper-layer step invoked after `tune run` succeeds.

This is the **foundation** rip: it introduces the wrapper layer that the wrapper-only policy hinges on. Land this first; the other rips (β γ δ ε ζ) only delete code from the recipes once this exists.

## Scope

- **Recipes affected**: `lora_finetune_single_device_stable`, `lora_finetune_single_device_nightly`, `lora_finetune_distributed_stable`
- **New file**: `src/tools/torchtune/post_finetune.py` (81 lines on the reference branch — `discover_epoch_dirs()` + `main()`, reuses the existing F2 helpers from `custom_recipe_utils.py`)
- **`custom_recipe_utils.py`**: F2 helpers stay (they get imported by `post_finetune.py`); only the recipe-side binding is removed
- **Template change**: `src/tools/torchtune/templates/finetune_template.slurm` gains a post-`tune run` step gated on `TUNE_EXIT_CODE`:
  ```bash
  if [ \$TUNE_EXIT_CODE -eq 0 ]; then
      python -m cruijff_kit.tools.torchtune.post_finetune --config finetune.yaml
  fi
  ```
- **Tests**: F2 helpers remain covered by `tests/unit/test_adapter_config_rewrite.py` (no new tests required, but verify the existing ones still pass against the wrapper-imported helpers)

## Reference implementation

[PR #523](https://github.com/niznik-dev/cruijff_kit/pull/523) (closed) contains the working diff. The branch `465-strict-wrapper-principle` can be cherry-picked from — relevant commit: \`1026916\`.

## Acceptance criteria

- [ ] \`grep -n \"cruijff_kit patch\" src/tools/torchtune/custom_recipes/*.py\` returns no F2-related markers in any of the three recipes
- [ ] \`src/tools/torchtune/post_finetune.py\` exists and is invoked from the SLURM template
- [ ] \`pytest tests/unit/\` passes (focus on \`test_adapter_config_rewrite.py\`)
- [ ] **SLURM smoke**: scaffold + run a 1B fine-tune end-to-end. Confirm \`post_finetune.py\` rewrites \`adapter_config.json\` and a downstream eval cell can load the resulting adapter cleanly.

—MxC

niznik-dev:
Closes #524.

## Summary

First rip in the per-rip carve-up of closed PR #523, under milestone [**Wrapper-only principle (#465)**](https://github.com/niznik-dev/cruijff_kit/milestone/11). Moves the adapter-housekeeping step (rewrite `adapter_config.json`'s base path on adapter-only saves, or stash adapter files on merged saves) **out of** the recipes' `save_checkpoint` hook and **into** a new wrapper-layer script that the SLURM template invokes after `tune run` succeeds.

This rip is intentionally narrow:

- F1 (`epochs_to_save`), F3 (custom metrics), F4 (`num_workers`), F5 (`tqdm_miniters`), F6 (dead `setup_logger` import) — all stay in the recipes for now. Each gets its own follow-up PR per the carve-up plan (issues #525–#529).
- The `# !--- cruijff_kit patch ---!` blocks that *previously* mixed F1 and F2 are now F1-only — the import block, the init block, and the post-save block.
- F2 helpers (`rewrite_adapter_config_base_path`, `stash_adapter_files`) stay in `custom_recipe_utils.py`; they're now imported by `post_finetune.py` instead of the recipes.

## New dependencies

None. \`post_finetune.py\` uses \`yaml\` (already a dep) and reuses the existing F2 helpers from \`custom_recipe_utils.py\`.

## Test plan

- [x] \`ruff check\` + \`ruff format --check\` on changed Python files (clean)
- [x] \`pytest tests/unit/\` — 1190 tests pass
- [x] \`python -m py_compile\` clean on all three recipes + \`post_finetune.py\`
- [ ] **SLURM smoke**: scaffold + run a 1B fine-tune end-to-end. Confirm the SLURM-level post-finetune step rewrites \`adapter_config.json\` and a downstream eval cell can load the resulting adapter cleanly.

—MxC


niznik-dev:
@msalganik I'm adding you as reviewer for each of these. Also, I'm experimenting with Marin to see if she's helpful.

src/tools/torchtune/post_finetune.py
```diff
@@ -0,0 +1,81 @@
...
+    parser.add_argument(
+        "--config",
+        default="finetune.yaml",
+        help="Path to the finetune.yaml the recipe used (defaults to cwd/finetune.yaml)",
+    )
+    args = parser.parse_args()
+
+    cfg_path = Path(args.config)
+    with cfg_path.open() as f:
+        cfg = yaml.safe_load(f)
```
niznik-dev: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

This re-reads `finetune.yaml` from scratch. The recipe doesn't hand state to this wrapper — they run as separate processes. The yaml is the only contract between them. Anything the wrapper needs to know about the run (output dir, base model path, save mode) has to live in the yaml.

src/tools/torchtune/post_finetune.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if not output_dir.is_dir():
+        raise SystemExit(f"output_dir from {cfg_path} not found: {output_dir}")
+
+    epochs = discover_epoch_dirs(output_dir)
+    if not epochs:
+        log.warning(f"No epoch_N subdirs in {output_dir}; nothing to do.")
+        return
+
+    for epoch in epochs:
```
niznik-dev, line 85: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

Walks every `epoch_N/` directory the recipe produced. Why every? Because users can save multiple epochs today (that's F1 — `epochs_to_save`, still in the recipe for now). Each saved epoch is a self-contained checkpoint dir that needs its own housekeeping.

Once β (#525) lands and selective saving goes away, every epoch saves by default. The loop shape here doesn't need to change — it was already written for the multi-epoch case.

src/tools/torchtune/post_finetune.py
```diff
@@ -0,0 +1,81 @@
...
+        if save_adapter_weights_only:
```
niznik-dev, line 86: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

📍 **look here**

The two save modes need genuinely different post-processing — this is the most domain-knowledge-dense spot in the file:

- **Adapter-only** (`save_adapter_weights_only: True`) — rewrite `adapter_config.json`'s `base_model_name_or_path` to a local absolute path. Otherwise the saved adapter only loads on the machine that wrote it, because PEFT's auto-load tries to reach the path that was current at training time.
- **Merged** (`save_adapter_weights_only: False`) — stash the adapter files into a subdir. PEFT auto-detection is greedy: if it sees adapter files next to a merged checkpoint, it'll load adapter+base instead of the merged file. Moving the adapter files out of sight prevents that.

If you've ever wondered "why does my merged checkpoint sometimes behave like an adapter" — this is the spot that fixes it.

src/tools/torchtune/custom_recipes/lora_finetune_single_device_stable.py
```diff
@@ -16,10 +16,8 @@
 from omegaconf import DictConfig, ListConfig
 
 # !--- cruijff_kit patch ---!
```
niznik-dev, line 18: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

The patch-block header rename is worth a beat: this block used to import *both* F1 (`validate_epochs_to_save`) and F2 (`rewrite_adapter_config_base_path`, `stash_adapter_files`). With F2 gone, it's an F1-only block — hence the rename to "epochs_to_save - selective checkpoint saving." The header now tracks what's actually in the block.

You'll see the identical edit in `lora_finetune_single_device_nightly.py` and `lora_finetune_distributed_stable.py` — same three surgical changes (import, init, save-checkpoint hook), three times.

niznik-dev (commented):
🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

Two-minute tour for anyone landing on this PR cold. This is the first of six per-rip carve-outs from closed PR #523, deliberately narrow: it moves *one* feature — adapter housekeeping — out of the recipes and into a wrapper script. F1, F3, F4, F5, F6 are still in the recipes; each gets its own follow-up under milestone [Wrapper-only principle (#465)](https://github.com/niznik-dev/cruijff_kit/milestone/11).

**Where to look, in flow order:**

1. `src/tools/torchtune/templates/finetune_template.slurm` (+6 lines) — the new SLURM hook that runs *after* `tune run` exits.
2. `src/tools/torchtune/post_finetune.py` (new, 81 lines) — the wrapper script that hook invokes.
3. The three recipes — each gets the same three surgical edits (import block, init, save-checkpoint hook). Walking through `lora_finetune_single_device_stable.py` is enough; the other two are identical edits applied to identical structure.

**What did NOT move:**

📍 The F2 helper functions (`rewrite_adapter_config_base_path`, `stash_adapter_files`) still live in `src/tools/torchtune/custom_recipes/custom_recipe_utils.py`. They're now imported by `post_finetune.py` instead of by the recipes. If you grep for them you'll still find them — that's intentional.

**Why this shape:**

🟢 The #465 audit found that adapter housekeeping is fundamentally a post-save *file* operation — it doesn't need training-loop state. So it doesn't belong inside `train()`; it belongs after the recipe exits cleanly. Moving it lets the recipes stay closer to upstream torchtune.

A few inline notes on the spots most worth slowing down at.

src/tools/torchtune/custom_recipes/lora_finetune_single_device_stable.py
niznik-dev: This makes sense to me - we're ripping out two different things incrementally that had a shared import

src/tools/torchtune/post_finetune.py
```diff
@@ -0,0 +1,81 @@
+"""Post-finetune adapter housekeeping (#465 wrapper-principle move).
+
+Reads the run's finetune.yaml to determine the output_dir, base checkpoint
+dir, and adapter-save mode; then walks every `epoch_N` subdirectory and
```
niznik-dev, line 4: I think this is why we wanted to put it inside of torchtune originally (not having to walk through a bunch of output folders) but it should be trivial and will now comply with our principles

src/tools/torchtune/templates/finetune_template.slurm
niznik-dev, line 1: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

This is the SLURM batch script template — `setup_finetune.py` materializes one of these per run, and SLURM executes it. The bulk of the file (lines 1–36) is unchanged: SBATCH headers, conda setup, GPU monitoring, `tune run`. This PR adds 6 lines at the bottom: a new post-train step that invokes the wrapper script `post_finetune.py`.

Two design choices worth knowing about the new step:

- **Gated on training success** — if `tune run` exits non-zero, the housekeeping is skipped. No point rewriting adapter configs for a run that didn't finish.
- **Doesn't override `TUNE_EXIT_CODE`** — the SLURM job's exit code stays equal to the training exit code, regardless of whether housekeeping itself fails. The wrapper is advisory; the recipe stays the source of truth on whether the run succeeded.

src/tools/torchtune/post_finetune.py
niznik-dev, line 1: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

Brand new file (81 lines). This is the wrapper script that `finetune_template.slurm` invokes after `tune run` exits cleanly. It does three things:

1. Reads `finetune.yaml` from disk to learn the output dir, base model path, and save mode.
2. Walks every `epoch_N/` subdirectory the recipe wrote.
3. Applies adapter housekeeping to each — either rewriting `adapter_config.json` (adapter-only saves) or stashing adapter files (merged saves).

📍 The actual file-mutation logic (`rewrite_adapter_config_base_path` and `stash_adapter_files`) wasn't written for this PR — those helpers have been in `custom_recipe_utils.py` for a while. This file just makes them callable as a standalone post-training step instead of being patched into the recipe's `save_checkpoint` hook.

See the inline notes below for the spots inside this file most worth slowing down on.

src/tools/torchtune/custom_recipes/lora_finetune_single_device_stable.py
niznik-dev, line 1: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

One of three custom torchtune recipes cruijff_kit maintains — the "stable" single-device LoRA variant (torchtune's stable release line, no multi-GPU). It's a vendored fork of torchtune's upstream recipe with cruijff_kit-specific patches marked in `# !--- cruijff_kit patch ---!` blocks.

Three surgical edits in this PR:

1. **Import block** — removed two helper imports (`rewrite_adapter_config_base_path`, `stash_adapter_files`); `validate_epochs_to_save` stays. Patch header renamed since the block no longer mixes features.
2. **`__init__`** — removed the `self._base_model_path` assignment. The wrapper reads it from yaml directly now.
3. **`train()` save-checkpoint hook** — removed the post-save adapter housekeeping call. The wrapper does it after the recipe exits.

The same three edits land identically in `lora_finetune_single_device_nightly.py` and `lora_finetune_distributed_stable.py` — file-level notes there are brief since the pattern repeats.

src/tools/torchtune/custom_recipes/lora_finetune_single_device_nightly.py
niznik-dev, line 1: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

The "nightly" single-device LoRA recipe — structural twin to `lora_finetune_single_device_stable.py`, but tracking torchtune's nightly builds instead of stable releases. cruijff_kit maintains both so users can opt into newer torchtune features without a global pin change.

The three surgical edits in this PR are **identical** to the stable recipe: import block trim, `_base_model_path` removal, save-checkpoint hook trim. If you've read through the stable diff, this one is the same edits again. Line numbers differ slightly because upstream nightly carries a few extra features.

src/tools/torchtune/custom_recipes/lora_finetune_distributed_stable.py
niznik-dev, line 1: 🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

🟢 **note**

The multi-GPU/distributed variant of the LoRA recipe. Same family as the other two custom recipes; the differences are all about distributed-training plumbing (DDP setup, rank-aware file writes).

The three surgical edits in this PR are functionally identical to the single-device recipes, with one structural quirk worth pointing out:

📍 The F2 call sites used to live inside an `if self._is_rank_zero:` block — only rank 0 wrote adapter files, since the other ranks were holding identical state. That `is_rank_zero` guard is gone now along with everything else. The wrapper runs as a single post-train Python process after distributed has already torn down, so rank-awareness moves out of the recipe and becomes a non-issue at the wrapper layer.

src/tools/torchtune/templates/finetune_template.slurm
```diff
@@ -35,4 +35,10 @@ TUNE_EXIT_CODE=$?
 kill $GPU_MONITOR_PID 2>/dev/null
 wait $GPU_MONITOR_PID 2>/dev/null
 
+# Post-finetune adapter housekeeping (#465 wrapper-principle).
```
niznik-dev: We need to remove this - too much info

src/tools/torchtune/templates/finetune_template.slurm
```diff
@@ -35,4 +35,10 @@ TUNE_EXIT_CODE=$?
...
+# Runs only if training succeeded; failures here don't override TUNE_EXIT_CODE.
```
niznik-dev: We need to remove this - too much info

src/tools/torchtune/post_finetune.py
```diff
@@ -0,0 +1,81 @@
+"""Post-finetune adapter housekeeping (#465 wrapper-principle move).
```
niznik-dev: Don't need to reference 465

src/tools/torchtune/post_finetune.py
```diff
@@ -0,0 +1,81 @@
...
+
+import yaml
+from torchtune import utils
+
+from cruijff_kit.tools.torchtune.custom_recipes.custom_recipe_utils import (
+    rewrite_adapter_config_base_path,
+    stash_adapter_files,
+)
+
+log = utils.get_logger("DEBUG")
```
niznik-dev, line 30: Let's do a trial run to see if the skill is logging as well

src/tools/torchtune/post_finetune.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+
+import argparse
+import re
+from pathlib import Path
+
+import yaml
+from torchtune import utils
+
+from cruijff_kit.tools.torchtune.custom_recipes.custom_recipe_utils import (
```
niznik-dev: These are probably in the wrong location now

msalganik:
We discussed this. Looks good to me, pending minor changes we discussed. I don't need to see it again before merge. But happy to talk if it would be helpful.

Issue #10, opened 2025-04-16
Add torchtune upgrade checklist to repo


niznik-dev:
Probably in a README.md-like file somewhere. Here's my general outline:

* Create an upgrade issue and branch (e.g. 44-upgrade-to-torchtune-0_7_0)
* Checkout the branch locally with the ttenv activated and run `pip install -U torchtune==x.y.z`
* When successful, run any tests we have (#5?)
  * If there are problems, update code until they pass
  * Then make sure the code is still compatible with the previous version
* Merge into main

msalganik:
@niznik-dev this looks great to me.  One suggestion would be to not put it in the readme.  I think of the readme as for them.  This feels more like a checklist for us.  I've sometimes put these things in the wiki.  

Or maybe we could have a readme for users and a readme for developers?

niznik-dev:
We're on the same page, either a dev readme or the Wiki, wasn't sure if you had used the Wiki before 😁

niznik-dev:
This is now here: https://github.com/niznik-dev/predicting-zygosity/wiki/Dependency-Upgrade-Checklists

niznik-dev:
@msalganik Can you confirm the link in the comment above is what you had in mind? If so I'll close this issue.

msalganik:
Yes.  Looks great @niznik-dev.  My hope is that this can be a living document as we learn from each upgrade that we do.  Given that we are in a fast moving ecosystem we need to get good at upgrades.

msalganik:
Also, @niznik-dev I closed the issue, but if you wanted to do that let me know.  I'm still not sure on the workflow we want.

niznik-dev:
I think ideally the assignee should close but I'm not against it, especially for something this simple.

PR #533, opened 2026-05-28
report_generator: scorer-aware heading + footnote for continuous-only experiments

### LINKED ISSUE #519: report_generator: Risk Metrics heading is wrong for continuous_scorer-only experiments
## Background

v0.3.2 ships continuous_scorer (#508), the first non-`risk_scorer` supplementary metric. `report_generator.generate_report()` was written when the only supplementary metrics were risk-scorer outputs (ECE, Brier, AUC), so it hardcodes both:

- the section heading: `## Risk Metrics`
- the explanatory footnote: `C-ECE (Confidence ECE): "when I'm X% confident, am I right X% of the time?" R-ECE (Risk ECE): "when I say P(Y=1)=X, is Y=1 X% of the time?"`

Both render under continuous-only experiments and are misleading there.

## Reproduce

The v0.3.2 smoke test produced this report:

`/scratch/gpfs/MSALGANIK/niznik/ck-projects/folktexts/v032_smoke_acs_age_2026-05-21/analysis/report.md`

The Risk Metrics section there contains mae / rmse / r_squared / parse_rate, with the C-ECE footnote underneath.

## Suggested fix

In `src/tools/inspect/report_generator.py`, branch on whether any risk-scorer metrics are present:

- If only continuous_scorer metrics are present: emit `## Regression Metrics`, drop the C-ECE/R-ECE footnote, and add a short \"mae/rmse: lower is better; r_squared: higher is better; parse_rate: fraction of parseable numeric outputs\" note instead.
- If both risk and continuous are present (uncommon but possible): emit two sub-sections, one for each.
- If only risk metrics: keep current behavior.

The numbers in the table itself are already correct — this is a labeling-only change.

## Why it's not a v0.3.2 blocker

The data renders correctly; only the section label is wrong. continuous_scorer users won't be confused by their own metric names, just by the heading above them. Patch-bump candidate.

—MxC

niznik-dev:
Closes #519

# Description

\`report_generator.generate_report()\` hardcoded \`## Risk Metrics\` + the C-ECE/R-ECE explanation, so \`continuous_scorer\`-only experiments (added in v0.3.2 via #508) rendered mae / rmse / r_squared / parse_rate under a misleading risk-metrics label with a calibration footnote about binary-probabilistic predictions. The v0.3.2 ACS-age smoke test surfaced it cleanly.

This PR makes the heading + footnote scorer-aware:

| Category present | Heading | Footnote |
|---|---|---|
| risk-only | \`## Risk Metrics\` | C-ECE / R-ECE / Brier / AUC (unchanged) |
| continuous-only | \`## Regression Metrics\` | mae / rmse / r_squared / parse_rate |
| mixed (rare) | Two sub-sections: \`## Risk Metrics\` then \`## Regression Metrics\` | One footnote per sub-section |

Numbers in the tables were always correct — this is a labeling-only fix per the issue body.

## Implementation

- New module-level constants: \`CONTINUOUS_METRIC_SUFFIXES = {\"mae\", \"rmse\", \"r_squared\", \"parse_rate\"}\`, \`_RISK_FOOTNOTE\`, \`_CONTINUOUS_FOOTNOTE\`.
- New helpers: \`_is_continuous_metric(name)\` classifies by suffix after the last \`/\` (matching the \`{scorer_name}/{metric_name}\` format already used by \`detect_metrics\`); \`_split_calibration_by_category(results)\` returns \`(risk_results, continuous_results)\` with each list carrying only the relevant metric subset.
- The calibration-only path now builds a \`sections\` list and loops; the mixed case gets two table-with-footnote sub-sections under one Executive Summary.
- The accuracy-present path (\`_format_model_table\`) checks which categories are in \`supp_names\` and emits the matching footnote(s).

## New Dependencies

None.

# Testing Instructions

- \`pytest tests/unit/test_report_generator.py\` — 54 passed (48 existing risk-only / format / report tests unchanged, 6 new for continuous-only and mixed paths)
- \`pytest tests/unit/\` — 1196 passed, no regressions
- \`ruff check\` + \`ruff format --check\` — clean

New test coverage:

\`\`\`
TestContinuousOnlyReport
  test_regression_heading_not_risk_metrics
  test_continuous_footnote_present_risk_footnote_absent
  test_metric_values_render

TestMixedRiskAndContinuousReport
  test_both_headings_present_in_order
  test_both_footnotes_present
  test_metrics_split_across_tables
\`\`\`

To verify against a real v0.3.2-shaped report: re-run report_generator on the v0.3.2 ACS-age smoke artifacts at \`/scratch/gpfs/MSALGANIK/niznik/ck-projects/folktexts/v032_smoke_acs_age_2026-05-21/\` and confirm the \`## Risk Metrics\` heading is now \`## Regression Metrics\` and the C-ECE footnote is gone.

—MxC


niznik-dev (commented):
🎭 **Skeptical Code Reviewer** 🎭 — [**Bram**](https://github.com/niznik-dev/llm-personas/blob/main/personas/bram/bram.md)

Look, Mattie. The fix works. Lines up with what #519 asked for, tests are green, I'm not arguing with that.

But you're pulling up empty pots and acting surprised there's no crabs.

The discriminator is `CONTINUOUS_METRIC_SUFFIXES = {"mae", "rmse", "r_squared", "parse_rate"}`. Hardcoded. In `report_generator`. Sitting forty modules away from `continuous_scorer.py`, which is where those metric names actually originate — `@scorer(metrics=[mae(), rmse(), r_squared(), parse_rate()])`. Two registries. Both have to know the same thing. Add `mape` to the scorer tomorrow and forget the frozenset, this exact bug recurs — silently, no test catches it, you find out when someone runs an analysis report and the C-ECE footnote shows up under `mape`. We've been here. We just did this dance.

The scorer already tells you what it is. The supplementary metric name is literally `continuous_scorer/mae`. The *prefix* discriminates. You're parsing the suffix and matching against a set when `name.startswith("continuous_scorer/")` would have done the job without anyone having to keep two lists in sync. Or — better — `continuous_scorer.py` exports `METRIC_NAMES = frozenset({"mae", "rmse", "r_squared", "parse_rate"})` and the report importer pulls from there. Decoration uses the same constant. Adding a metric is one edit. Forgetting is impossible because the file you edit is the file that breaks.

That's the one that matters. Couple of smaller things while I'm here.

**The helpers reference `CalibrationResult` and the class is defined forty lines later.** Forward refs work, sure. Reads backwards. Move the helpers below the dataclass or move the dataclass above the helpers — doesn't matter which, but pick one. Right now a reader hits `_split_calibration_by_category(results: list["CalibrationResult"])` before they know what a CalibrationResult is.

**The `cont_footnotes` conditional is doing two things at once.** `[sample_footnote] if sample_footnote and not risk_results else []` — that's gating on whether there's a sample footnote AND whether the previous section already emitted one. Worked it out, but it took a beat. Either pull it apart:

```python
already_shown = bool(risk_results)
cont_footnotes = [_CONTINUOUS_FOOTNOTE]
if sample_footnote and not already_shown:
    cont_footnotes.insert(0, sample_footnote)
```

or just always put the sample footnote on the first section regardless. Either way, the dedup logic shouldn't require a second read.

**CHANGELOG calls this "labeling-only."** It isn't, anymore — you added two-sub-section emission for the mixed case. That's a structural change. Worth a half-sentence flagging it, because someone scanning the changelog for a risk-only experiment needs to know the rendering path changed even if their numbers didn't.

One thing you got right. `test_metrics_split_across_tables` actually verifies that AUC doesn't leak into the regression section — checks the substance, not just the heading text. That's the test I want to see. If you'd skipped it I'd be more annoyed.

What I'd do: move the metric set into `continuous_scorer.py`, import it in `report_generator`, and write a test that asserts the imported set matches what the scorer's `@scorer(metrics=[...])` decoration actually produces. That way adding a continuous metric forces both registrations to update — or fails loudly. The fix as-is will silently break the next time someone extends the scorer and forgets the frozenset.

Hm. Get the registry right and this is solid work.

niznik-dev (commented):
🎭 **Skeptical Code Reviewer** 🎭 — [**Bram**](https://github.com/niznik-dev/llm-personas/blob/main/personas/bram/bram.md)

Hm. That's actually not bad.

When I said "hoist the constant and add a sync-check test," what I was after was for you to close the drift channel. You went one better — `_METRICS` is a single list, the `@scorer` registration and `METRIC_NAMES` are both derived from it. No sync to check because there's nothing to sync. That's the right move; the test I asked for doesn't need to exist.

The *"names are explicit because inspect-ai clobbers `__name__`"* comment matters. Without it the next reader will try to "simplify" to `m.__name__` and find every metric classified as `metric_wrapper`. Documenting the constraint is what separates a one-time fix from a fix that survives a reader. Keep that comment exactly as is.

One follow-up while it's fresh: `viz_helpers.has_risk_scorer` is the same drift pattern on the risk side — `"auc_score" in m` substring match against a name declared in `risk_scorer.py`. Worth a separate issue. Not this PR.

Ship it.

niznik-dev (commented):

Different milestone from the #465 series — this one's [Quiet Bugs II (#10)](https://github.com/niznik-dev/cruijff_kit/milestone/10) territory. A label-and-footnote fix for `report_generator` so `continuous_scorer`-only experiments stop rendering under a `## Risk Metrics` heading with a C-ECE/R-ECE explanation underneath.

📍 **Two commits, in conversation.** First commit is the original fix Mattie wrote against #519. Second commit is a refactor Bram nudged: the metric-name set used to classify columns originally lived in `report_generator`, forty modules away from `continuous_scorer.py` where the metrics are declared. The second commit moves that set into `continuous_scorer.py` and has `report_generator` import it, so adding a new continuous metric auto-classifies. Reading the PR commit-by-commit shows the structural decision more clearly than the merged diff.

**Reading order I'd suggest:**

1. `continuous_scorer.py` — see the `_METRICS = [(name, factory), ...]` pattern. This is the foundation of the second commit. Inline 📍 below.
2. `report_generator.py` — where the labeling fix actually lives. Inline 📍 at the new import.
3. `test_report_generator.py` — six new tests, two classes. The substance check (`test_metrics_split_across_tables`) is the one to slow down on.
4. `CHANGELOG.md` — one new Unreleased / Fixed entry.

---

### Per-file overview

| File | What this PR does to it | Skim/Read/Pause |
|---|---|---|
| `src/tools/inspect/scorers/continuous_scorer.py` | Adds `_METRICS` list + exported `METRIC_NAMES` frozenset. The `@scorer` registration is now derived from `_METRICS` instead of literal calls. Names are explicit because inspect-ai's `@metric` clobbers `__name__`. | **read** |
| `src/tools/inspect/report_generator.py` | Imports `METRIC_NAMES` from `continuous_scorer`, removes the local duplicate. The actual #519 fix lives here: `generate_report` and `_format_model_table` branch heading + footnote on whether metrics are continuous, risk, or both. Mixed case emits two ordered sub-sections. | **read** |
| `tests/unit/test_report_generator.py` | Six new tests across two classes — `TestContinuousOnlyReport` (3) and `TestMixedRiskAndContinuousReport` (3). The mixed-case tests verify metrics don't leak across sub-sections, not just that the headings appear. | skim |
| `CHANGELOG.md` | One Unreleased / Fixed entry. | skim |

---

### Severity callouts

🟢 **note — review-cycle shape.** The two-commit history (original fix + Bram-driven refactor) is the kind of PR shape worth keeping. The Option A discussion didn't bloat the diff — net change went from "one duplicate constant in report_generator" to "single source of truth + cross-module import." The refactor cost about ten minutes of review-driven structure and closes a whole drift channel.

🟡 **watch — scorer modules now have downstream consumers.** Before this PR, `continuous_scorer.py` was imported only by inspect-ai task wiring. It now also exports a name set that `report_generator` consumes. Renaming the module or removing `METRIC_NAMES` would break the report path. Worth knowing if `scorers/` ever gets reorganized.

💭 **follow-up.** Bram flagged `viz_helpers.has_risk_scorer` (line 208–210) as the same drift pattern on the risk side — `"auc_score" in m` substring matching against a name declared in `calibration_metrics.py`. Tracking that as a separate Quiet Bugs II issue rather than expanding scope here.

---

Mattie did the rip; I'm just walking the path. Trust your own read.

src/tools/inspect/scorers/continuous_scorer.py
```diff
@@ -110,7 +110,24 @@ def compute(scores: list[Score]) -> float:
     return compute
 
 
-@scorer(metrics=[mae(), rmse(), r_squared(), parse_rate()])
+# Source of truth for this scorer's metrics. Names are explicit because
+# inspect-ai's @metric wraps each factory and clobbers __name__ (all four
+# would report as "metric_wrapper"), so introspection can't recover them.
+# Adding one: define the @metric factory above, then append (name, factory)
+# here — the @scorer registration and METRIC_NAMES are derived from this list.
+_METRICS: list[tuple[str, callable]] = [
```
niznik-dev, line 118: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

📍 **look here.** The `(name, factory)` tuple format is non-obvious. It looks like it could be simplified to `_METRICS = [mae, rmse, r_squared, parse_rate]` plus `METRIC_NAMES = frozenset(m.__name__ for m in _METRICS)`. It can't: inspect-ai's `@metric` decorator wraps each factory and clobbers `__name__`, so all four would report as `metric_wrapper`. The comment above this block documents the constraint — Bram specifically noted it should stay, because a future reader might be tempted to "clean up" otherwise.

src/tools/inspect/report_generator.py
```diff
@@ -17,15 +17,66 @@
...
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from datetime import datetime
 from pathlib import Path
 from typing import Optional
 
 import pandas as pd
 
+from cruijff_kit.tools.inspect.scorers.continuous_scorer import (
+    METRIC_NAMES as _CONTINUOUS_METRIC_NAMES,
```
niznik-dev, line 28: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

📍 **look here.** This is the line that makes the drift-fix structural rather than nominal. `report_generator` no longer carries its own list of continuous-metric names — it imports the canonical set from where the scorer declares it. Adding `mape` to `continuous_scorer._METRICS` tomorrow automatically updates this classifier with no second touchpoint, which was the entire point of the second commit.

src/tools/inspect/scorers/continuous_scorer.py
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**Overview — read.** Adds the `_METRICS` list and exports the `METRIC_NAMES` frozenset; the `@scorer` registration is now *derived* from `_METRICS` rather than from four literal metric calls. This is the foundation the second commit rests on.

Pointer: the `(name, factory)` tuple shape is the load-bearing bit — the inline 📍 below explains why the names are spelled out instead of read from `__name__`.

src/tools/inspect/report_generator.py
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**Overview — read.** Where the #519 fix actually lives. Imports `METRIC_NAMES` from `continuous_scorer` (dropping the local duplicate set) and branches the heading + footnote on whether an experiment's metrics are continuous, risk, or both.

Pointer: the mixed case emits two ordered sub-sections — that's the new rendering path. The import line is flagged with an inline 📍 below.

tests/unit/test_report_generator.py
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**Overview — skim.** Six new tests across two classes — `TestContinuousOnlyReport` (3) and `TestMixedRiskAndContinuousReport` (3).

Pointer: if you slow down anywhere, make it `test_metrics_split_across_tables`. It checks that metrics don't *leak* across sub-sections, not just that the headings render — the substance, not the label.

CHANGELOG.md
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**Overview — skim.** One Unreleased / Fixed entry for the #519 heading fix.

Pointer: the rendering path changed for the mixed case (two sub-sections now), not just the labeling — worth a glance to confirm the entry reflects that. Bram has the detailed take in the top-level review.

niznik-dev:
Merging as discussed on Friday afternoon

PR #536, opened 2026-05-29
δ (F5) — Remove tqdm_miniters override from recipes (#527)

### LINKED ISSUE #527: δ (F5) — Remove tqdm_miniters override from recipes ("see how it goes" empirical item)
Part of #465 (wrapper-only principle). Carved out of closed PR #523.

## Context

The custom recipes patch the tqdm progress-bar instantiation in `train()` to pass `miniters=cfg.get(\"tqdm_miniters\", 100)`. The original motivation was readability of SLURM logs (tqdm flushes far less often, producing one line per N steps instead of many per second). The #465 audit flagged this as a low-risk patch — but a patch nonetheless, sitting inside `train()`'s loop body where the wrapper-only policy says nothing of ours should live.

This is the **\"see how it goes\"** item Matt explicitly flagged in PR #523's test plan. The empirical question is whether removing the override breaks Claude's ability to read SLURM logs (tqdm's default behavior is much chattier). If it does, we may want to address it via a wrapper-level log filter rather than reintroducing the in-`train()` patch.

## Scope

- **Recipes affected**: `lora_finetune_single_device_stable`, `lora_finetune_single_device_nightly`, `lora_finetune_distributed_stable` (~5 lines each — `__init__` config-read + tqdm-instantiation patch site)
- **YAML refs affected**: drop \`tqdm_miniters\` from \`finetune_template.yaml\` / \`1B_defaults.yaml\` / \`8B_defaults.yaml\` if present
- **\`setup_finetune.py\`**: drop argparse / merge entries for \`tqdm_miniters\` if present
- **Tests affected**: any \`test_tqdm_miniters_*\` cases
- **User-facing config flags removed**: \`tqdm_miniters\`

## Reference implementation

[PR #523](https://github.com/niznik-dev/cruijff_kit/pull/523) (closed) contains the working diff. The branch `465-strict-wrapper-principle` can be cherry-picked from.

## Acceptance criteria

- [ ] \`grep -n \"cruijff_kit patch\" src/tools/torchtune/custom_recipes/*.py\` returns no F5-related markers
- [ ] \`grep -rn tqdm_miniters src/ tests/\` returns nothing
- [ ] \`pytest tests/unit/\` passes
- [ ] **Real SLURM run + Claude observation**: launch a 1B fine-tune. Confirm Claude can read the resulting SLURM log without crashing (the original concern). If it crashes or struggles, **do not merge** — open a follow-up to handle log-volume at the wrapper layer (e.g., a tail/grep filter applied to the log before Claude reads it).

—MxC

niznik-dev:
Closes #527

# Description

Fifth rip of milestone [Wrapper-only principle (#465)](https://github.com/niznik-dev/cruijff_kit/milestone/11) — removes the `tqdm_miniters` progress-bar override. The three recipes' `pbar = tqdm(...)` lines revert to upstream torchtune exactly; the `__init__` config-read and the `tqdm_miniters` yaml keys are dropped.

| Recipe | Restored line |
|---|---|
| single_device (stable + nightly) | `pbar = tqdm(total=self._steps_per_epoch)` |
| distributed (stable) | `pbar = tqdm(total=self._steps_per_epoch, disable=not (self.rank == 0))` |

## Why this is safe — #527's "see how it goes" question, answered honestly

**#527's acceptance criterion is readability: "confirm Claude can read the resulting SLURM log without crashing."** Not byte count.

The override existed to keep tqdm output short enough that Claude wouldn't choke when reading the log. **That failure mode no longer occurs** — the Read tool now paginates large / carriage-return-heavy logs (offset/limit), so they read fine in slices. Verified directly: read a **626 KB / 10,000-step** chatty log (100-epoch 1B fine-tune, override removed) through the Read tool — clean, no crash. With the original constraint gone, the override is obsolete and reverts to upstream per the wrapper-only principle.

**Honest accounting of the tradeoff** (controlled A/B — same config, seed, data; only variable = presence of `miniters=100`):

| Arm | progress-bar lines / epoch |
|---|---|
| treatment (no `miniters`) | ~51 |
| control (`miniters=100`) | ~2 |

So removal *does* make the bar ~25× more verbose **at fast step-speed** (~12 it/s, below tqdm's 0.1 s `mininterval`). But verbose ≠ unreadable, and readability is the criterion. (Earlier drafts of this PR claimed the override was a "no-op" — that was wrong; it was suppressing ~96% of bar lines. The correct justification is obsolescence of the crash mode, not absence of effect.)

## New Dependencies

None.

# Testing Instructions

- `grep -rn tqdm_miniters src/ tests/ docs/` → empty
- `grep -n "pbar = tqdm" src/tools/torchtune/custom_recipes/*.py` → matches upstream torchtune exactly
- `ruff check` + `ruff format --check` on the 3 recipes — clean
- `python -m py_compile` clean on all three recipes
- `pytest tests/unit/` — 1160 passed
- **SLURM acceptance (#527 criterion):** a 1B fine-tune trains cleanly on the edited recipe, and the resulting chatty 626 KB / 10k-step log reads cleanly via the Read tool — no crash. A/B control run (`miniters=100`, identical config) quantifies the verbosity delta above.

—MxC


niznik-dev:
🎻 **MxC** ✳️

**SLURM acceptance run ✅** — the open box from the testing section.

Launched a 1B stable-recipe fine-tune on the δ branch (`words_5L_80P_1000`, batch 8, 1 epoch → 100 steps, job `8958147`):

- **Trains cleanly without the override** — loss 1.71 → 0.037, run completed, F2 adapter housekeeping fired (`Rewrote adapter_config.json base_model_name_or_path`).
- **Log readable** — 14 KB, 52 training-bar refreshes over 100 steps; parsed without trouble.
- **Bonus confirmation of the thesis**: this run was fast (~11.7 it/s ≈ 0.085 s/step, *below* tqdm's 0.1 s `mininterval`), so the bar coalesced to ~1 refresh per 2 steps — fewer than one-per-step. Volume is governed by `mininterval`/step-time, not the deleted `miniters`. Paired with the 246 KB / 2428-refresh real-pace bench log (also readable), both the fast and chatty ends are covered.

Calling the "see how it goes" question settled: removing the override leaves log volume unchanged (or lower) and readability intact.

—MxC

niznik-dev:
🎻 **MxC** ✳️

**Correction + controlled A/B** — supersedes my earlier "volume unchanged" comment, which was wrong.

I ran a proper A/B: identical 100-epoch 1B config (same seed/data, `num_workers=0` both), the only variable being `miniters=100`.

| Arm | bytes | lines | tqdm refreshes (CR) | bar lines/epoch |
|---|---|---|---|---|
| no `miniters` (this branch) | 626,046 | 3,603 | 5,201 | ~51 |
| `miniters=100` (control) | 136,968 | 1,154 | 300 | ~2 |

So the override was **not** a no-op — it suppressed ~96% of progress-bar lines. **But that was never the acceptance criterion.** #527 asks whether Claude can *read* the log without crashing. I read the chatty 626 KB / 10k-step log directly via the Read tool (it paginates large / `\r`-heavy files) — clean, no crash.

**Verdict:** the override mitigated a failure mode (choking on chatty logs) that no longer exists. It's obsolete, not ineffective, so it reverts to upstream per the wrapper-only principle. Logs are more verbose without it but fully readable. PR description + CHANGELOG updated to say this accurately.

—MxC

niznik-dev:
🧭 [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md) 🧭

Let's walk through this one, Mattie. It's a clean revert — the recipes give back two lines each to upstream, and three config files drop one key. Nothing here changes training behavior; it changes how chatty the progress bar is.

| File | What | Skim / Read / Pause | Pointer |
|---|---|---|---|
| `lora_finetune_single_device_stable.py` | `__init__` tqdm read + pbar override removed | **read** | the `pbar = tqdm(...)` line |
| `lora_finetune_single_device_nightly.py` | same revert as stable | skim | mirrors stable |
| `lora_finetune_distributed_stable.py` | same revert; pbar keeps the rank guard | **read** | `disable=not (self.rank == 0)` |
| `finetune_template.yaml` | drops `tqdm_miniters: 100` | skim | — |
| `1B_defaults.yaml` / `8B_defaults.yaml` | drops `tqdm_miniters: 100` | skim | — |
| `CHANGELOG.md` | Removed entry, #527 | skim | — |

📍 **look here** — the restored pbar lines match upstream torchtune exactly: bare `tqdm(total=self._steps_per_epoch)` on single_device, and the rank-guarded form on distributed. That's the thing to confirm, and it holds.

🟢 **note** — the A/B in the thread above does the honest accounting: removing the override makes the bar ~25× more verbose at fast step-speed, but the log stays readable. The justification is obsolescence (paginated reading), not "no effect." Good that the PR text was corrected to say so.

🟡 **watch** — the live SLURM run exercised the **stable single_device** recipe only. The nightly and distributed pbar edits are verified by inspection against upstream (they match), not by a run. Low risk for a one-line revert, but the multi-GPU path is unexercised here.

🟡 **watch** — this PR, #537, and #538 all edit the same three recipe files in different regions. Whichever merges last may want a trivial rebase.

🟢 **note** — every file here is a pure deletion/revert, so I've folded the per-file trail-markers into the table above rather than repeating them inline. Say the word if you want them anchored in the diff.

Issue #254, opened 2026-01-15
run-experiment: Auto-retry with reduced batch_size on OOM


niznik-dev:
## Summary

When fine-tuning jobs fail with OOM errors, `run-experiment` should automatically retry with a reduced batch_size rather than requiring manual intervention.

## Context

During the ACS mobility format comparison experiment, we had to manually:
1. Detect OOM failures
2. Cancel jobs
3. Edit `finetune.yaml` to reduce batch_size (e.g., 128 → 96 → 64)
4. Resubmit

This happened multiple times across different runs, with the pattern being predictable based on model size and prompt format (verbose prompts need smaller batches).

## Proposed Behavior

When `run-experiment` detects an OOM failure:
1. Parse the error from slurm output
2. Reduce batch_size by some factor (e.g., ÷2 or -25%)
3. Update the config
4. Resubmit automatically
5. Log the adjustment

### Configuration Options
- `max_oom_retries`: How many times to retry (default: 3)
- `batch_size_reduction`: Strategy for reduction (e.g., `halve`, `subtract_16`, `multiply_0.75`)
- `min_batch_size`: Floor to avoid infinite retries (default: 8)

## Bonus: Proactive Scaling

Could also learn from prior runs in the same experiment. If 1B_verbose needed batch_size 64, automatically start 3B_verbose at 32.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #540, opened 2026-06-01
Hoist risk scorer METRIC_NAMES into canonical module

### LINKED ISSUE #534: viz_helpers.has_risk_scorer: hoist METRIC_NAMES into the canonical scorer module (drift parallel to #519)
## Background

`viz_helpers.has_risk_scorer` (`src/tools/inspect/viz_helpers.py:208-210`) classifies eval dataframes by checking for the substring `"auc_score"` in supplementary metric names:

```python
@property
def has_risk_scorer(self) -> bool:
    """True when the evaluation used risk_scorer (has AUC metric)."""
    return any("auc_score" in m for m in self.supplementary)
```

The name `"auc_score"` is declared in `src/tools/inspect/scorers/calibration_metrics.py` (the `@metric` factory) and consumed by `risk_scorer.py` / `numeric_risk_scorer.py`. Two registries — the substring constant in `viz_helpers` and the metric name in `calibration_metrics` — that have to stay in sync. Same drift-prone pattern that PR #533 fixed for `continuous_scorer` (#519).

## Why this matters

Bram surfaced this during the #533 review: the registry-drift critique applies equally to the risk side. It won't fire today — `risk_scorer` ships with AUC and isn't slated to drop it — but the bug-shape recurs the next time anyone renames or reorganizes calibration metrics, and the failure mode is silent (`has_risk_scorer` returns False when it shouldn't, downstream plot routing changes silently).

## Suggested fix

Mirror PR #533's pattern:

1. In `src/tools/inspect/scorers/risk_scorer.py` (or `calibration_metrics.py`, wherever the canonical declaration feels right), add a `METRIC_NAMES` frozenset alongside the metric factories — derived from a single `_METRICS = [(name, factory), ...]` list per #533.
2. `viz_helpers.has_risk_scorer` imports it and consumes via membership check on the suffix-after-last-`/`:
   ```python
   return any(m.rsplit("/", 1)[-1] in RISK_METRIC_NAMES for m in self.supplementary)
   ```
3. Decide: should `numeric_risk_scorer` share the same `METRIC_NAMES` (it appears to publish the same suite) or export its own? Either is fine; document the choice.

## Reference

PR #533 (#519 fix), commit `721f09d` — `_METRICS` + `METRIC_NAMES` pattern in `continuous_scorer.py`. The "names are explicit because inspect-ai's @metric clobbers `__name__`" comment is load-bearing; carry it through.

## Why it's not a v0.3.3 blocker

`has_risk_scorer` works today and produces correct routing for the metric names the risk scorers actually ship. Patch-bump-worthy hygiene, not a regression-fix priority.

—MxC

niznik-dev:
Closes #534

# Description

Mirrors the `continuous_scorer` pattern (PR #533) for the risk side, removing the registry-drift hazard between `viz_helpers.has_risk_scorer` and the metric declarations.

- **`risk_scorer.py`** — adds a single canonical `_METRICS` list (source of truth) and derives both the `@scorer(metrics=...)` registration and an exported `METRIC_NAMES` frozenset from it.
- **`numeric_risk_scorer.py`** — reuses the same `_METRICS` (it publishes the identical 5-metric suite) instead of re-declaring the import block + metric list. Net −13 lines.
- **`viz_helpers.py`** — `has_risk_scorer` now classifies by membership of the suffix-after-`/` against `METRIC_NAMES`, replacing the `"auc_score"` substring scan.

**Decisions** (the issue asked us to make + document them):
- *Canonical home:* `risk_scorer.py` — where the `@scorer` lives, already imports the calibration metrics, defines `mean_risk_score`. Closest mirror to `continuous_scorer`.
- *Sharing:* `numeric_risk_scorer` shares `risk_scorer`'s `_METRICS` rather than exporting its own, since the suites are identical.

**Behavior note:** `has_risk_scorer` is now slightly more sensitive — it fires on any of the 5 risk metric names, not just `auc_score`. Since both risk scorers always ship the full suite together, this never changes real-world routing; it just makes detection robust to a future AUC drop (the point of the issue).

### Bonus: code-comment convention
While mirroring the pattern, I was about to propagate the `(#519)`/`(#534)` style already present in the scorer modules. Added a `## Code Conventions → Code Comments` section to `CLAUDE.md` banning issue/PR numbers in code comments (provenance belongs in commits/PRs, recoverable via `git blame`), and scrubbed the four existing `(#519)` references.

> Out of scope, flagged for a follow-up: one pre-existing `PR A (#254)` comment remains in `src/tools/run/_submit_common.py:363` (different subsystem).

## New Dependencies

None.

# Testing Instructions

```bash
ruff check src/tools/inspect/scorers/*.py src/tools/inspect/viz_helpers.py src/tools/inspect/report_generator.py
ruff format --check src/tools/inspect/scorers/*.py src/tools/inspect/viz_helpers.py src/tools/inspect/report_generator.py
python -m pytest tests/unit/ -k "risk or viz or report or scorer or continuous or metric" -q
```

All 338 selected unit tests pass; lint + format clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,6 +109,17 @@ All use Llama-3.2-1B-Instruct with `{ck_data_dir}/capitalization/words_5L_80P_10
 
 **Purpose:** Catch regressions in skills, ensure documentation changes don't break workflows, validate end-to-end integration.
 
+## Code Conventions
+
+### Code Comments
+
+- **No issue/PR numbers in code comments.** Comments must explain the code as it
+  stands, self-contained. Provenance ("why did this change?") belongs in the
+  commit message and PR description — `git blame` recovers it when needed.
+  - ❌ `# Exported so viz_helpers (#534) can classify columns without re-listing.`
+  - ✅ `# Exported so viz_helpers can classify columns without re-listing.`
+  - Pointers to code that still exists are fine (e.g. `# See src/tools/.../inspect_task.py`) — those reference live structure, not ephemeral tracker state.
+
 ## Git Workflow
 
 ### Issue Management
--- a/src/tools/inspect/report_generator.py
+++ b/src/tools/inspect/report_generator.py
@@ -32,7 +32,7 @@ from cruijff_kit.tools.inspect.viz_helpers import detect_metrics, display_name
 # Supplementary metric names have the form "{scorer_name}/{metric_name}" (e.g.
 # "risk_scorer_cruijff_kit/auc_score", "continuous_scorer/mae"); we classify
 # by the part after the last "/" against the set the continuous_scorer module
-# itself exports, so adding a new metric there auto-classifies here (#519).
+# itself exports, so adding a new metric there auto-classifies here.
 
 _RISK_FOOTNOTE = (
     '*C-ECE (Confidence ECE): "when I\'m X% confident, am I right X% of the time?" '
@@ -526,7 +526,7 @@ def _format_model_table(
     if calibration:
         # Emit only the footnote(s) matching the metric categories actually
         # present, so a continuous-only experiment doesn't render the
-        # misleading C-ECE/R-ECE explanation (#519).
+        # misleading C-ECE/R-ECE explanation.
         has_risk = any(not _is_continuous_metric(n) for n in supp_names)
         has_continuous = any(_is_continuous_metric(n) for n in supp_names)
         if has_risk:
@@ -830,7 +830,7 @@ def generate_report(
     calibration_only = not metrics and bool(calibration_results)
 
     # sections is a list of (heading, table_markdown, footnotes) tuples so the
-    # mixed case (#519) can emit a Risk Metrics sub-section followed by a
+    # mixed case can emit a Risk Metrics sub-section followed by a
     # Regression Metrics sub-section under the same Executive Summary.
     sections: list[tuple[str, str, list[str]]] = []
 
--- a/src/tools/inspect/scorers/continuous_scorer.py
+++ b/src/tools/inspect/scorers/continuous_scorer.py
@@ -122,7 +122,7 @@ _METRICS: list[tuple[str, callable]] = [
     ("parse_rate", parse_rate),
 ]
 
-# Exported so report_generator (#519) can classify columns by scorer category
+# Exported so report_generator can classify columns by scorer category
 # without re-listing names.
 METRIC_NAMES: frozenset[str] = frozenset(name for name, _ in _METRICS)
 
--- a/src/tools/inspect/scorers/numeric_risk_scorer.py
+++ b/src/tools/inspect/scorers/numeric_risk_scorer.py
@@ -12,13 +12,10 @@ predicted label is the positive (last) label, otherwise the negative (first).
 from inspect_ai.scorer import scorer, Score, CORRECT, INCORRECT
 from inspect_ai.solver import TaskState
 from inspect_ai.scorer import Target
-from .risk_scorer import mean_risk_score
-from .calibration_metrics import (
-    expected_calibration_error,
-    risk_calibration_error,
-    brier_score,
-    auc_score,
-)
+
+# numeric_risk_scorer publishes the identical metric suite as risk_scorer, so it
+# reuses the same canonical _METRICS list rather than re-declaring it.
+from .risk_scorer import _METRICS
 
 
 def _parse_risk_score(text: str) -> float | None:
@@ -36,15 +33,7 @@ def _parse_risk_score(text: str) -> float | None:
     return value
 
 
-@scorer(
-    metrics=[
-        mean_risk_score(),
-        expected_calibration_error(),
-        risk_calibration_error(),
-        brier_score(),
-        auc_score(),
-    ]
-)
+@scorer(metrics=[factory() for _, factory in _METRICS])
 def numeric_risk_scorer(labels: tuple[str, str] = ("0", "1")):
     """
     Scorer that parses a numeric probability from the model's text output.
--- a/src/tools/inspect/scorers/numeric_risk_scorer.py
+++ b/src/tools/inspect/scorers/numeric_risk_scorer.py
@@ -14,8 +14,8 @@ from inspect_ai.solver import TaskState
 from inspect_ai.scorer import Target
 
 # numeric_risk_scorer publishes the identical metric suite as risk_scorer, so it
-# reuses the same canonical _METRICS list rather than re-declaring it.
-from .risk_scorer import _METRICS
+# reuses risk_scorer's public suite helper rather than re-declaring the metrics.
+from .risk_scorer import risk_metric_suite
 
 
 def _parse_risk_score(text: str) -> float | None:
@@ -33,7 +33,7 @@ def _parse_risk_score(text: str) -> float | None:
     return value
 
 
-@scorer(metrics=[factory() for _, factory in _METRICS])
+@scorer(metrics=risk_metric_suite())
 def numeric_risk_scorer(labels: tuple[str, str] = ("0", "1")):
     """
     Scorer that parses a numeric probability from the model's text output.
--- a/src/tools/inspect/scorers/risk_scorer.py
+++ b/src/tools/inspect/scorers/risk_scorer.py
@@ -39,7 +39,6 @@ def mean_risk_score() -> Metric:
 # would report as "metric_wrapper"), so introspection can't recover them.
 # Adding one: define/import the @metric factory, then append (name, factory)
 # here — the @scorer registration and METRIC_NAMES are derived from this list.
-# numeric_risk_scorer imports this same list, so the two stay in lockstep.
 _METRICS: list[tuple[str, callable]] = [
     ("mean_risk_score", mean_risk_score),
     ("expected_calibration_error", expected_calibration_error),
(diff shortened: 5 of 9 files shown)


CLAUDE.md
```diff
@@ -109,6 +109,17 @@ All use Llama-3.2-1B-Instruct with `{ck_data_dir}/capitalization/words_5L_80P_10
 
 **Purpose:** Catch regressions in skills, ensure documentation changes don't break workflows, validate end-to-end integration.
 
+## Code Conventions
```
niznik-dev, line 112: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**CLAUDE.md** — *skim.* Pure doc. New `Code Comments` convention banning issue/PR numbers in code comments — provenance lives in commits/PRs, recoverable via `git blame`. No behavior.

niznik-dev (commented):

Let's walk through this one, Mattie. Six files, but only one is real work — `risk_scorer.py` becomes the source of truth; the rest are consumers and comment hygiene. Tour is ordered by flow, not filename.

| File | What it does in this change | Route | Pointer |
|---|---|---|---|
| `scorers/risk_scorer.py` | Adds canonical `_METRICS` + exported `METRIC_NAMES`; `@scorer` derives from it | **read** | the new block above `@scorer` |
| `scorers/numeric_risk_scorer.py` | Drops duplicated metric imports/list; reuses `risk_scorer._METRICS` | **read** | the import swap + one-line `@scorer` |
| `viz_helpers.py` | `has_risk_scorer` matches metric *names* instead of an `"auc_score"` substring | **pause** | the property body — behavior shifts here |
| `report_generator.py` | Comment scrub only — three `(#519)` tags removed | **skim** | no code change |
| `scorers/continuous_scorer.py` | Comment scrub only — one `(#519)` tag removed | **skim** | no code change |
| `CLAUDE.md` | New "Code Comments" convention banning issue numbers in code | **skim** | the new subsection |

No 🟠 concern, no 🔴 block. Clean mirror of a pattern that already earned its keep on the continuous side. The one thing I'd keep an eye on is the sklearn-at-import coupling in `viz_helpers` — flagged inline as 🟡. Per-file trail markers are anchored on each file below.

src/tools/inspect/report_generator.py
```diff
@@ -32,7 +32,7 @@
 # Supplementary metric names have the form "{scorer_name}/{metric_name}" (e.g.
 # "risk_scorer_cruijff_kit/auc_score", "continuous_scorer/mae"); we classify
 # by the part after the last "/" against the set the continuous_scorer module
-# itself exports, so adding a new metric there auto-classifies here (#519).
+# itself exports, so adding a new metric there auto-classifies here.
```
niznik-dev, line 35: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**report_generator.py** — *skim.* Comment scrub only: three `(#519)` tags removed across the file (here, L529, L833). The classification logic is untouched.

src/tools/inspect/scorers/continuous_scorer.py
```diff
@@ -122,7 +122,7 @@ def compute(scores: list[Score]) -> float:
     ("parse_rate", parse_rate),
 ]
 
-# Exported so report_generator (#519) can classify columns by scorer category
+# Exported so report_generator can classify columns by scorer category
```
niznik-dev, line 125: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**continuous_scorer.py** — *skim.* Comment scrub only: one `(#519)` tag removed. The `_METRICS` / `METRIC_NAMES` pattern this PR mirrors onto the risk side already lives here — worth a glance for the parallel.

src/tools/inspect/scorers/numeric_risk_scorer.py
```diff
@@ -12,13 +12,10 @@
...
 from inspect_ai.scorer import Target
-from .risk_scorer import mean_risk_score
-from .calibration_metrics import (
-    expected_calibration_error,
-    risk_calibration_error,
-    brier_score,
-    auc_score,
-)
+
+# numeric_risk_scorer publishes the identical metric suite as risk_scorer, so it
```
niznik-dev, line 16: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**numeric_risk_scorer.py** — *read.* Drops the duplicated metric imports + re-listed suite; reuses `risk_scorer._METRICS`. Net -13 lines. The interesting bit is this import swap and the one-line `@scorer` just below.

---

💭 **question** — `_METRICS` is underscore-private but imported across the module boundary. It's documented and intentional, and a public re-export would be more ceremony than this small package needs. No change requested — just naming the tension so nobody "fixes" it later by accident.

src/tools/inspect/scorers/risk_scorer.py
```diff
@@ -34,15 +34,26 @@ def compute(scores: list[Score]) -> float:
...
-@scorer(
-    metrics=[
-        mean_risk_score(),
-        expected_calibration_error(),
-        risk_calibration_error(),
-        brier_score(),
-        auc_score(),
-    ]
-)
+# Source of truth for this scorer's metrics. Names are explicit because
```
niznik-dev, line 37: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**risk_scorer.py** — *read.* The source of truth. A `_METRICS` list of `(name, factory)` pairs that the `@scorer` registration *and* the exported `METRIC_NAMES` both derive from. The load-bearing comment about `@metric` clobbering `__name__` carried over from the continuous side — that's the explanation a future reader needs for why the names are hand-written instead of introspected.

---

🟢 **note** — `list[tuple[str, callable]]` uses the builtin `callable` rather than `typing.Callable`. It mirrors `continuous_scorer` exactly, so consistency wins — flagging only so the parallel reads as deliberate, not a thing to "correct" on one side and split the two modules.

src/tools/inspect/viz_helpers.py
```diff
@@ -32,6 +32,8 @@
     EvalTask,
 )
 
+from .scorers.risk_scorer import METRIC_NAMES as _RISK_METRIC_NAMES
```
niznik-dev, line 35: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**viz_helpers.py** — *pause.* The one file where behavior shifts. `has_risk_scorer` moves from substring-match to name-membership (see the property body below).

---

🟡 **watch** — This adds a module-load-time import of `risk_scorer`, which transitively pulls in `calibration_metrics` → `sklearn`. `viz_helpers` previously had no sklearn dependency at import time. Anything importing `viz_helpers` for its lightweight helpers (`display_name`, `deduplicate_eval_files`) now drags sklearn along on import. Not a problem today — both already ride together in the analyze path — but it's the kind of coupling that becomes one if a leaner consumer ever imports this module.

src/tools/inspect/viz_helpers.py
```diff
@@ -206,8 +208,15 @@ class DetectedMetrics:
 
     @property
     def has_risk_scorer(self) -> bool:
-        """True when the evaluation used risk_scorer (has AUC metric)."""
-        return any("auc_score" in m for m in self.supplementary)
+        """True when the evaluation used risk_scorer/numeric_risk_scorer.
```
niznik-dev, line 211: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

🟢 **note** — Detection is now broader: it fires on *any* of the five risk metric names, where the old code only saw `auc_score`. Since both risk scorers always ship the full suite, real-world routing is unchanged — this only hardens against a future where AUC is dropped. Recorded on the line so it's not a surprise.

src/tools/inspect/scorers/numeric_risk_scorer.py
niznik-dev: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

🟢 **resolved** in 61d71b4. The cross-module private import is gone: `risk_scorer` now exposes a public `risk_metric_suite()` that both scorers call, and `_METRICS` stays module-private (matching `continuous_scorer`). The underscore contract holds again, and the metric-instantiation comprehension lives in one place.

niznik-dev:
🎻 **MxC** ✳️

**Changes since the original PR description** — this grew during review, so here's the current shape for a pre-merge once-over. (The opening description still reflects the first commit only.)

| Commit | What | Why |
|---|---|---|
| `f1400c5` | Hoist `_METRICS`/`METRIC_NAMES` into `risk_scorer.py`; `viz_helpers.has_risk_scorer` matches by name; `numeric_risk_scorer` reuses the canonical source. Plus the "no issue numbers in code comments" convention in `CLAUDE.md`. | The original #534 fix. |
| `61d71b4` | Replace the cross-module `from .risk_scorer import _METRICS` with a public `risk_metric_suite()` helper both scorers call. | Review feedback: importing an underscore-private name across modules breaks the "internal to this module" contract — a later refactor of `risk_scorer` could rename/inline `_METRICS` and silently break `numeric_risk_scorer`. The helper is the intentional shared surface; `_METRICS` stays private (matching `continuous_scorer`). |
| `d05f70f` | New tests (338 → 360): metric-name **contract** (crosses the inspect-ai boundary), the `detect_metrics → has_risk_scorer` **pipeline** on real-shaped columns, and `has_risk_scorer` parametrized over every real metric name. | The pre-existing tests passed *coincidentally* (on a synthetic `"ece"` name) and didn't exercise the behavior change. The internal "`METRIC_NAMES` == names-of-`_METRICS`" check is a tautology; only crossing into inspect-ai's registry proves our hand-written names match what it actually emits. |
| `0ee4b03` | Inline comment at the `registry_info` import. | Documents the dependency below. |

**⚠️ One thing to eyeball before merge** — `tests/unit/scorers/test_risk_scorer.py` imports from a **private** inspect-ai path:

```python
from inspect_ai._util.registry import registry_info
```

This is the only way to read the name a `@metric` is registered under (which is what lands in the eval-log column), so it's load-bearing for the contract test. Test-only; fails *loudly at import* if a version bump moves it, rather than passing silently. Flagging it explicitly because reaching into `_util` is a judgment call you should get to veto, not discover.

Everything green: ruff check + format clean, 360 unit tests pass.

—MxC

niznik-dev:
61d71b4 in particular is a much cleaner implementation so I'm glad we did that one

Issue #473, opened 2026-05-07
estimate_compute: predict wall-time from token throughput rather than sample count


niznik-dev:
# Describe the Feature

`estimate_compute` currently predicts wall-time by scaling a prior run's wall-time by the ratio of sample counts (N). This breaks when sequence length changes between the prior and new run — per-step time depends on `seq_len`, not just N (e.g. ~3× per-step slowdown at k=200 vs k=25 on `lostmiddle_kxp_3B`).

Proposed fix: capture **tokens-per-second-per-GPU** (`tps/gpu`) from completed runs and predict new wall-time as `total_tokens_new / tps_gpu_prior`. Because `total_tokens = N × seq_len`, this naturally absorbs both N and seq_len changes.

# Status

> **Note from @niznik-dev**: Preserving this proposal as a follow-up — the **basic** resource estimation (PR #349, shipped in v0.3.0) is now in main, and this is the more accurate variant Matt sketched on top of it. I want to think through it more before committing to a scope or timeline; this issue is the home for that thinking.

# Background

Follow-up to [#243](https://github.com/niznik-dev/cruijff_kit/issues/243) (closed by PR #349). Originally proposed by @msalganik in [this 2026-05-04 comment](https://github.com/niznik-dev/cruijff_kit/issues/243#issuecomment-4371855312) with a detailed implementation sketch.

# Implementation outline (from Matt's sketch, paths normalized post-#441)

| File | Action | Change |
|---|---|---|
| `src/tools/slurm/throughput_parsers.py` | NEW (~80 lines) | `parse_torchtune_throughput(slurm_out)` and `parse_inspect_throughput(slurm_out)`. Hard-fail on missing throughput line so format drift is visible. |
| `src/tools/slurm/compute_summary.py` | EXTEND | `build_summary` writes 4 new tps fields per job. |
| `src/tools/slurm/estimate_compute.py` | EXTEND | Replace `scale_finetune_time` / `scale_eval_time` with throughput-based equivalents using prior `tps_gpu_*`. `recommend_batch_size` and `estimate_from_prior` signatures unchanged. |
| `src/tools/torchtune/custom_recipes/lora_finetune_*_stable.py` | EXTEND | Add a `[ck-tps]` print at end of training — format-stable anchor we own (hedges against torchtune wandb format drift). |
[...]

Issue #451, opened 2026-05-03
Make run-experiment's execution logs deterministic so analyze-experiment compute section doesn't silently skip


msalganik:
## For Mattie

**Problem.** The handoff between `run-experiment` and `analyze-experiment` is implicit and fragile. `analyze-experiment`'s compute-utilization section (step 6b) harvests SLURM job IDs via the regex `Result: Job ID (\d+)` against `{experiment_dir}/logs/run-torchtune.log` and `{experiment_dir}/logs/run-inspect.log`. If those logs don't exist — or exist but don't carry the canonical structured action lines — the compute section silently skips, per its own docs ("Skip silently if no run logs exist").

`run-experiment` is supposed to produce both execution logs. In practice it doesn't, reliably. The skill's execution path is documented as prose under `optimizers/torchtune/*.md` and `evaluators/inspect/*.md` — a list of steps for the agent to perform, with each step expected to also emit the right log line as a side-effect. There is no callable script. When the agent (correctly!) does the operational work via raw `sbatch` from the `Bash` tool, structured log lines aren't written. Result: `run-experiment.log` exists with prose narrative, but `run-torchtune.log` / `run-inspect.log` don't exist, and `analyze-experiment` silently produces a report with no compute section.

**Why fix it.** Just got bitten on `consistency_curve_nth14_k20`. All the raw data was on disk (`gpu_metrics.csv` per run, sacct/seff history for all 15 jobs), but `analyze-experiment` couldn't find the job IDs and skipped compute analysis. I wrote a one-off `_run_perf_review.py` (~200 lines) that re-implemented exactly what `analyze-experiment` step 6b already does. Same compute helpers (`parse_seff_output`, `summarize_gpu_metrics`, `format_compute_table` from `cruijff_kit.tools.slurm.compute_metrics`) — I just had to find the JIDs myself.

This will keep happening on every experiment until the contract becomes deterministic. The silent-skip behavior also actively hides the failure: someone who didn't realize compute analysis was supposed to be there would never know they're missing it. That makes the bug self-reinforcing — no one notices, no one files an issue, the gap persists.

**What changes after the fix.**
- `run-experiment` always emits both `logs/run-torchtune.log` and `logs/run-inspect.log` with canonical structured entries (`SUBMIT_JOB` / `SUBMIT_EVAL` action types containing `Result: Job ID <N> submitted at <ts>` lines), regardless of whether the agent took the prose-execution path or a script-execution path.
[...]

Issue #447, opened 2026-05-02
Make --training_samples actually slice the training dataset


msalganik:
## For Mattie

**Problem.** `setup_finetune.py` exposes a `--training_samples` flag (settable via `setup_finetune.yaml`), but it doesn't actually slice the dataset — today it only prints a warning if your step count looks low. The flag never reaches the generated `finetune.yaml`, and the dataset loader has no slice parameter to honor anyway. So setting `training_samples: 100` in a config does nothing to the actual fine-tune.

**Why fix it.** For consistency-curve experiments (vary N from 50 → 1000, eval on a fixed val set), we currently have to pre-generate 5 separate JSON files. With this fix we generate **one** large dataset once and slice with `training_samples` per run, while every run evaluates against the same large val set. Tighter, comparable CIs at small N; consistency tests get cheap to re-run.

**What changes after the fix.**
- `training_samples: 100` in `setup_finetune.yaml` → only the first 100 rows of `train` are loaded for fine-tuning that run.
- `dataset_val` is intentionally **unchanged** — every run sees the full val split, so val accuracy is comparable across N.
- Existing experiments that don't set `training_samples` are unaffected (no behavior change when the field is absent).
- Existing step-count guard / warnings still fire as before.
- Two new regression tests prevent this bug from returning, both in the standard `pytest tests/unit/` CI step.

**Acceptance.** Two fine-tune runs pointed at the same shared JSON, with `training_samples: 50` and `training_samples: 1000` in their respective `setup_finetune.yaml`s, produce `finetune.yaml`s that:
1. Differ only in `dataset.max_samples` (val sections identical).
2. Show step counts proportional to N in their SLURM logs (50-row run has roughly 1/20 the steps of the 1000-row run).
3. Both regression tests pass: `pytest tests/unit/test_setup_finetune_main.py tests/unit/test_chat_completion_dataset.py tests/unit/test_text_completion_dataset.py`.

**Effort estimate.** ~15 lines of source + ~70 lines of tests across 5 files. Small.

---

### Investigation already done

Don't re-investigate — go straight to the patch. Key findings from tracing the chain:

- `src/tools/torchtune/setup_finetune.py:20` — `META_ARGS = ["training_samples"]` excludes it from YAML output.
- `src/tools/torchtune/setup_finetune.py:739-741` — the args→YAML loop skips anything in `META_ARGS`, so `training_samples` is never written to `finetune.yaml`.
[...]

PR #542, opened 2026-06-01
Sweep issue-number references out of src/ comments

### LINKED ISSUE #541: Sweep issue-number references out of code comments (src/, decide on tests/)
🎻 **MxC** ✳️

## Background

PR #540 (#534) introduced a **Code Conventions → Code Comments** rule in `CLAUDE.md`: **no issue/PR numbers in code comments.** Provenance ("why did this change?") belongs in commit messages and PR descriptions, recoverable via `git blame`. The rule is **absolute** — even forward-pointers to open issues get reworded to describe the gap *without* the number (e.g. `# no val loop yet` rather than `# no val loop yet (issue #474)`). Only pointers to live *code* (`# See src/tools/.../inspect_task.py`) are exempt.

#540 scrubbed the inspect-side scorer references that the rule was born from. This issue tracks bringing the **rest of the codebase** into line.

> Note: the convention surfaced because the original `#534` work was about to *propagate* the `(#519)` style further. The codebase predates the rule, so this is a one-time catch-up sweep, not a regression fix.

## Inventory — `src/` (production code)

| File:line | Current comment | Flavor |
|---|---|---|
| `run/_submit_common.py:160` | `(issue #451 comment)` | provenance |
| `run/_submit_common.py:335` | `with multi-key strategies (see #254 PR B)` | cross-pointer |
| `run/_submit_common.py:363` | `No active escalation in PR A (#254)` | provenance |
| `run/_submit_common.py:676` | `(see #254 PR A). Pass no_retry=True...` | cross-pointer |
| `torchtune/setup_finetune.py:894,955` | `recipe has no val loop yet (issue #474)` | forward-pointer |
| `slurm/throughput_parsers.py:11,77` | `see issue #473 for multi-GPU follow-up` | forward-pointer |
| `slurm/estimate_compute.py:418,433` | `not yet wired in — issue #492` | forward-pointer |
| `slurm/estimate_compute.py:573` | `legacy summaries pre-#491` | provenance |
| `inspect/setup_inspect.py:31` | `Empirically (see issue #318 ...)` | reference/rationale |
| `experiment/archive_experiment.py:144` | `(#498). A run counts as complete...` | provenance |

Each gets reworded to convey the same *what/why* without the number. Forward-pointers keep the substance ("no val loop yet", "single-GPU only for now") and drop the `#NNN`.

## Open scope question — `tests/`

`tests/` uses issue numbers **pervasively**, almost all as regression-guard docstrings:

```python
"""Regression: text_completion_dataset must honor max_samples ... (#447)."""
```

These are arguably a different animal — they document *which bug a test guards*, which is real value, and tests aren't read for "current behavior" the way prod code is. **Decision needed before sweeping tests:**

- **(a)** In scope — reword to describe the guarded behavior without the number.
- **(b)** Out of scope — test docstrings may name the originating issue (carve tests out of the convention).
- **(c)** Middle — keep the *behavioral* description, drop bare `(#NNN)` tags but allow `issue #NNN` where it adds genuine traceability.

Recommend resolving (a/b/c) here, then the convention in `CLAUDE.md` gets a one-line clarification about test docstrings either way.

## Acceptance

- [ ] All `src/` comments in the inventory reworded (number-free), behavior/rationale preserved
- [ ] `grep -rnE "#[0-9]+" src/ --include="*.py"` returns only false positives (e.g. `rsplit("/", 1)`)
- [ ] `tests/` scope decided (a/b/c) and applied
- [ ] `CLAUDE.md` convention clarified re: tests if needed

—MxC

niznik-dev:
Closes #541

# Description

Removes all 15 issue/PR-number references from `src/` comments, per the Code Comments convention introduced in #540, and scopes the convention to **`src/` only — tests exempt**.

## The rule, finalized: "tests fine, src never"

After inventorying every inline reference across the codebase, the pattern was clear: in `src/` the numbers mostly rot, but in `tests/` the regression-guard docstrings (`"""Regression: ... (#447)"""`) document *which bug a test guards* — traceability worth more than the rot risk. So:

- `CLAUDE.md` now reads **"No issue/PR numbers in code comments under `src/`"** with an explicit **"Tests are exempt"** clause.

## src/ rewordings (15 across 7 files)

Each keeps the *what/why*, drops the number:

| Flavor | Example before → after |
|---|---|
| Forward-pointer | `no val loop yet (issue #474)` → `no val loop yet` |
| Forward-pointer | `See issue #473 for multi-GPU follow-up` → `Multi-GPU support is a planned follow-up` |
| Superseded behavior | `Pre-#451, analyze-experiment's compute step…` → `Previously, analyze-experiment's compute step…` |
| Provenance | `No active escalation in PR A (#254).` → `No active escalation.` |
| Reference | `Empirically (see issue #318 investigation comment),` → `Empirically,` |

Two edits touch text *inside* runtime strings (the `estimate_compute` cross-model warning, the `throughput_parsers` distributed-recipe error) — verified no test asserts on the removed substrings.

## New Dependencies

None.

# Testing Instructions

```bash
# src/ is clean (only false positives like rsplit("/", 1)):
grep -rnE "#[0-9]+" src/ --include="*.py" | grep -vE "rsplit|\[len|#!/"

ruff check src/ && ruff format --check src/
python -m pytest tests/ -k "estimate_compute or throughput or compute_metrics or setup_inspect or setup_finetune or archive or submit or run_experiment" -q
```

461 tests pass for the affected modules; ruff clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -113,12 +113,17 @@ All use Llama-3.2-1B-Instruct with `{ck_data_dir}/capitalization/words_5L_80P_10
 
 ### Code Comments
 
-- **No issue/PR numbers in code comments.** Comments must explain the code as it
-  stands, self-contained. Provenance ("why did this change?") belongs in the
-  commit message and PR description — `git blame` recovers it when needed.
+- **No issue/PR numbers in code comments under `src/`.** Production-code comments
+  must explain the code as it stands, self-contained. Provenance ("why did this
+  change?") belongs in the commit message and PR description — `git blame`
+  recovers it when needed.
   - ❌ `# Exported so viz_helpers (#534) can classify columns without re-listing.`
   - ✅ `# Exported so viz_helpers can classify columns without re-listing.`
   - Pointers to code that still exists are fine (e.g. `# See src/tools/.../inspect_task.py`) — those reference live structure, not ephemeral tracker state.
+  - **Tests are exempt.** Test files may cite the originating issue/PR — a
+    `"""Regression: ... (#447)"""` docstring documents *which bug the test guards*,
+    and the issue holds the repro and discussion. That traceability is worth more
+    in test code than the rot risk costs.
 
 ## Git Workflow
 
--- a/src/tools/experiment/archive_experiment.py
+++ b/src/tools/experiment/archive_experiment.py
@@ -140,8 +140,8 @@ def inventory_experiment(experiment_dir, output_dir_base):
             )
 
     # --- Check completeness ---
-    # Each run's eval/ holds one cell directory per (task, epoch) pair (issue
-    # #498). A run counts as complete when at least one cell has eval logs.
+    # Each run's eval/ holds one cell directory per (task, epoch) pair. A run
+    # counts as complete when at least one cell has eval logs.
     incomplete_runs = []
     for entry in eval_matrix:
         run_name = entry.get("run")
--- a/src/tools/inspect/setup_inspect.py
+++ b/src/tools/inspect/setup_inspect.py
@@ -28,8 +28,8 @@ from cruijff_kit.tools.torchtune.model_configs import MODEL_CONFIGS
 TEMPLATE_PATH = Path(__file__).parent / "templates" / "eval_template.slurm"
 
 # Default ceiling on samples per HF batch for inspect-ai's HF provider.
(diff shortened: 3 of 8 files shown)


niznik-dev:
@msalganik Merging this but if you have reservations about leaving issue numbers in the tests we can talk. Since tests often arise from particular bugfixes I think those are actually helpful to leave.

Issue #539, opened 2026-05-29
Spike: run → summarize (strict) → explore, retiring deterministic plot-selection


niznik-dev:
<img src="https://raw.githubusercontent.com/niznik-dev/llm-personas/main/fusions/mattie/mattie-icon.png" width="36" align="absmiddle"> 🎻 **MxC** ✳️

## Summary

Spike to rethink the post-run analysis flow. Today the workflow is `run → summarize → analyze`, where **analyze** carries a large deterministic machinery that tries to *infer which plots to make* from the experimental design. That inference is the hard part — and we've concluded it's effectively un-generalizable: every experiment wants slightly different figures, and no static variable→view table covers the space.

This issue proposes shifting to:

```
run → summarize (stricter, script-driven) → explore (optional, Claude-driven)
```

and prototyping `explore` to see whether handing the "which figures, what interpretation" decision to Claude per-experiment is better than maintaining a deterministic plot-selection engine.

**This is a design + prototype spike.** It may not lead to production code; the exploration itself is the deliverable.

## Motivation / current pain

The "decide which plots to make" logic lives in `analyze-experiment`'s variable→view inference table and the supporting Python. Surface area today:

| Asset | ~Lines | Role |
|---|---|---|
| `src/tools/inspect/report_generator.py` | ~1000 | Deterministic markdown report: metric tables, Wilson CIs, calibration/ROC, compute utilization |
| `src/tools/inspect/viz_helpers.py` | ~580 | inspect-viz pre-built view selection + dedup + dynamic metric detection |
| `.claude/skills/analyze-experiment/` | ~1190 (5 files) | Maps experimental variables → which inspect-viz views to generate |
| `.claude/skills/summarize-experiment/` | ~290 | Lightweight metric extraction → `summary.md` |

The pain is concentrated in the inference layer (variable→view mapping). Each new experiment shape (new scorer, new factor structure, new metric) pressures that table to grow or special-case.

## Proposed workflow

1. **`run`** — unchanged.
2. **`summarize` (stricter)** — pure scripts, no judgment. Produces the always-correct, always-present numeric backbone for *any* experiment.
3. **`explore` (new, optional)** — Claude reads the summary + raw eval data and decides, per-experiment, what's worth plotting and what the results *mean*. No fixed plot menu.

### The summarize / explore boundary

Decided principle:

> **If something is meaningful for *every* experiment → `summarize`. Otherwise → `explore`.**

[...]

Issue #520, opened 2026-05-21
Document continuous_scorer in references/scorers.md


niznik-dev:
## Background

v0.3.2 ships continuous_scorer (#508), which produces four regression metrics: mae, rmse, r_squared, parse_rate. The design-experiment scorers reference (`.claude/skills/design-experiment/references/scorers.md`) currently documents only `match`, `includes`, `risk_scorer`, and `numeric_risk_scorer` — continuous_scorer is missing.

## What to add

A new \"### continuous_scorer\" subsection under \"## Custom Scorers (src/tools/inspect/scorers/)\" mirroring the structure of the existing entries:

- Source path
- Parameters table (none — continuous_scorer takes no kwargs)
- Metrics produced (mae, rmse, r_squared, parse_rate, with one-line descriptions)
- Requirements (only max_tokens needs to be wide enough for the expected number of digits; logprobs not required)
- Typical usage (regression targets — age, income, hours worked, commute time)
- Minimal YAML example:

\`\`\`yaml
scorer:
  - name: \"continuous_scorer\"
\`\`\`

Optionally a short note that for base models on continuous tasks, an `assistant_prefix: \"Answer: \"` (or similar) is often required to lift parse_rate off zero — the smoke test on 1B-Instruct base hit parse_rate=0 without it. Same caveat already applies to risk_scorer on base models.

## Why it's not a v0.3.2 blocker

Pure documentation gap. The scorer ships and works; the CHANGELOG and PR #508 both describe it. Anyone who finds it via the registry will know the metric names from the implementation.

—MxC

Issue #509, opened 2026-05-19
Make summarize/analyze skills continuous-scorer aware


niznik-dev:
## Context

PR #508 adds `continuous_scorer` and ACS continuous tasks (age, income, hours, commute) for regression-style evals. The scorer works end-to-end (verified), but the post-eval tooling downstream of `run-experiment` is binary-shaped throughout — anyone running a continuous experiment will hit broken or misleading output when they reach `summarize-experiment` or `analyze-experiment`.

## What's broken / missing

- **`src/tools/inspect/summary_binary.py`** — builds a 0/1 confusion matrix; will produce nonsense (or error) for continuous outputs like `"62096.0"`.
- **`.claude/skills/summarize-experiment/SKILL.md`** — partially aware (mentions "for non-binary tasks, only accuracy is reported"), but accuracy is meaningless for regression. Doesn't know `mae` / `rmse` / `r_squared` / `parse_rate` exist.
- **`.claude/skills/analyze-experiment/SKILL.md`** — generates `scores_by_task.html` keyed on match accuracy; `roc_curves.png` / `calibration_curves.png` / `prediction_histogram.png` are all gated on `risk_scorer` (binary probabilistic). No regression-shaped plots.
- **`.claude/skills/design-experiment/SKILL.md`** — defaults to thinking about binary classification tasks; doesn't surface continuous as a first-class option.

## What needs to happen

1. Add a `summary_continuous.py` (or generalize `summary_binary.py` → `summary.py` with scorer-aware dispatch) that pulls `mae` / `rmse` / `r_squared` / `parse_rate` from eval logs.
2. Update `summarize-experiment` SKILL.md to detect the scorer type and call the right summary tool.
3. Update `analyze-experiment` SKILL.md to generate regression plots (predicted-vs-target scatter, residual histogram, per-model metric table) when `continuous_scorer` is present.
4. Update `design-experiment` SKILL.md so continuous targets are a documented option, not an undocumented one.

## Scope note

Likely splits into 2-3 PRs (summary tool first, then skills, then design-experiment polish). Not a single-PR effort.

niznik-dev:
Closing as the analyze portion is either deferred or won't do

PR #544, opened 2026-06-01
Make summarize-experiment continuous-scorer aware

niznik-dev:
Refs #509 (summarize portion only — see scope note)

# Description

Teaches the `summarize-experiment` skill to handle continuous/regression experiments scored by `continuous_scorer`, which the binary-shaped tooling currently mishandles.

## What's added

**`src/tools/inspect/summary_continuous.py`** — the regression companion to `summary_binary.py`. It reads `continuous_scorer`'s *own per-sample outputs* (`prediction` / `target_value` / `error` from `Score.metadata`, via `read_eval_log`) and the authoritative aggregate metrics from `results.scores` — so its numbers match the scorer exactly rather than re-parsing model text. Beyond `mae`/`rmse`/`r_squared`/`parse_rate`, it reports a **prediction / target / residual distribution glance** the aggregates hide.

> Why the glance earns its keep — real `acs_age` smoke output:
> ```
> MAE: 17.53   RMSE: 21.94   R²: -0.94   Parse rate: 100%
>              min     max    mean     std
> Prediction  24.0    55.0   27.9    8.16
> Target      18.0    81.0   42.3   15.77
> Residual   -56.0    26.0  -14.4   16.58
> ```
> `R² = -0.94` alone reads as "noise." The distributions show the truth: a systematic **downward bias** (residual mean −14) and a **compressed range** (pred max 55 vs target 81). A flat prediction std would likewise reveal a model emitting one constant.

**`.claude/skills/summarize-experiment/SKILL.md`** — dispatches on scorer type: `continuous_scorer` → `summary_continuous.py` (regression table: MAE/RMSE/R²/Parse Rate), binary → the existing `summary_binary.py` path. Accuracy is no longer reported for regression (it's meaningless there).

**`tests/unit/test_summary_continuous.py`** — 8 tests covering metric extraction, the distribution glance, the constant-prediction (std=0) sanity case, parse-failure counting, NaN→None, and the fallback parse_rate.

## Scope note (why `Refs`, not `Closes`)

#509 lists four work items; this PR is deliberately **summarize-only**:
- ✅ **summary tool** + ✅ **summarize SKILL.md** — this PR.
- ⏸️ **analyze-experiment** — intentionally deferred. Per the #539 spike, the deterministic analyze machinery is slated for replacement by a freeform "explore" skill, so investing in regression plots there now would likely be thrown away.
- ✅ **design-experiment docs** — already handled by #520 (merged).

Suggest leaving #509 open for the analyze bit (or closing it as superseded once #539 lands) — your call.

## New Dependencies

None.

# Testing Instructions

```bash
ruff check src/tools/inspect/summary_continuous.py tests/unit/test_summary_continuous.py
ruff format --check src/tools/inspect/summary_continuous.py tests/unit/test_summary_continuous.py
python -m pytest tests/unit/test_summary_continuous.py tests/unit/test_summary_binary.py -q

# Against a real continuous eval:
python -m cruijff_kit.tools.inspect.summary_continuous <path-to-acs_age.eval>
```

40 summary/parse tests pass; verified end-to-end against a real `acs_age` `continuous_scorer` log.


src/tools/inspect/summary_continuous.py
```diff
@@ -0,0 +1,223 @@
...
+
+import argparse
+import json
+import math
+import statistics
+from pathlib import Path
+
+# The scorer key under which continuous predictions live, in both the per-sample
+# Score.metadata and the aggregate results.scores entry.
+SCORER_NAME = "continuous_scorer"
```
niznik-dev, line 32: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**summary_continuous.py** — *read.* The new tool. Reads the scorer's own per-sample outputs (`prediction`/`target_value`/`error`) and authoritative aggregates from `results.scores`, so its numbers match the scorer rather than re-parsing. The value-add over the four metrics is the distribution glance in `_distribution` (a flat prediction std flags a constant-emitting model).

---

🟡 **watch** — `SCORER_NAME = "continuous_scorer"` is a hardcoded duplicate of the scorer's registry name. If the scorer is ever renamed, `sample.scores.get(SCORER_NAME)` quietly misses and the per-sample distributions go empty — no error, just a thinner glance. This is the same registry-drift shape #534/#540 just addressed for metric names, so flagging it for consistency. Not a block: the name is stable and #539 may replace this machinery. If it outlives #539, importing the name from the scorer module (the way `viz_helpers` now imports `METRIC_NAMES`) would close it.

src/tools/inspect/summary_continuous.py
```diff
@@ -0,0 +1,223 @@
...
+            "message": f"No samples in {path}",
+            "path": str(path),
+        }
+
+    # Authoritative aggregate metrics, straight from the scorer's results.
+    agg: dict = {}
+    scorer_name = SCORER_NAME
+    if log.results and log.results.scores:
+        score = next(
+            (s for s in log.results.scores if s.name == SCORER_NAME),
```
niznik-dev, line 96: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

📍 **look here** — the aggregate-metric fallback: `next((s for s in scores if s.name == SCORER_NAME), scores[0])`. Pairs with the watch above — the same hardcoded name is load-bearing here too. If `SCORER_NAME` doesn't match *and* the continuous scorer isn't `scores[0]`, this grabs a different scorer (e.g. `match`), whose `metrics` have no `mae`/`rmse`, so every headline metric returns `None` silently. Fine today (continuous runs ship `continuous_scorer` as the scorer); noting it so a future reader knows the name gates two code paths, not one.

.claude/skills/summarize-experiment/SKILL.md
```diff
@@ -86,10 +86,12 @@ For each COMPLETED evaluation:
    ```bash
    python -m cruijff_kit.tools.inspect.parse_eval_log {path}
    ```
-3. Parse JSON output for accuracy
+3. Parse JSON output for the scorer's metrics (the `scorer` field tells you the type)
 4. **Map to epoch using SLURM job names** (see below)
-5. For binary tasks, also run `summary_binary.py` to get balanced accuracy and F1
-6. Record: run_name, task, epoch, accuracy, balanced_accuracy, f1, samples
+5. **Dispatch on scorer type:**
+   - **Binary classification** (`match`/`exact_match` on 0/1 targets): also run `summary_binary.py` for balanced accuracy and F1 (below).
```
niznik-dev, line 92: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**SKILL.md** — *read.* Dispatch on scorer type: `continuous_scorer` → `summary_continuous.py` (regression table), binary → the existing `summary_binary.py`. Reads cleanly, and dropping accuracy for regression is the right call — it's meaningless there. The new "Computing Regression Metrics" subsection and the regression Evaluation Results table downstream mirror the binary section's shape, so the doc stays parallel.

🟢 **note** — The base-model `parse_rate=0` pointer ties back to `references/scorers.md` (the #520 doc), so the two docs reinforce each other. Nice.

tests/unit/test_summary_continuous.py
```diff
@@ -0,0 +1,150 @@
+"""Unit tests for cruijff_kit.tools.inspect.summary_continuous.
```
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**test_summary_continuous.py** — *skim.* 8 tests, mocking `read_eval_log` with `SimpleNamespace` fakes (matches `test_parse_eval_log.py`'s pattern).

🟢 **note** — Coverage hits the cases that matter: authoritative-vs-fallback `parse_rate`, NaN→None, parse-failure counting, and the constant-prediction `std=0` case that locks the headline value-add. That last one is the test I'd have most wanted to see here — it proves the glance does the thing it exists to do.

niznik-dev (commented):

Let's walk through this one, Mattie. Three files: one new tool that does the real work, one test file, and the skill doc that routes to the tool. Tour by flow — the tool first, then how the skill reaches it, then the tests that pin it.

| File | What it does in this change | Route | Pointer |
|---|---|---|---|
| `src/tools/inspect/summary_continuous.py` | New regression summary tool — reads `continuous_scorer`'s per-sample metadata + aggregate metrics | **read** | the `SCORER_NAME` coupling, and the distribution glance in `_distribution` |
| `.claude/skills/summarize-experiment/SKILL.md` | Dispatches on scorer type; adds the regression-metrics path + table | **read** | the new "Computing Regression Metrics" subsection |
| `tests/unit/test_summary_continuous.py` | 8 tests, mock `read_eval_log` | **skim** | the constant-prediction (std=0) case |

One 🟡 **watch**, no 🟠/🔴 — this is a clean, well-tested addition. The watch is a coupling worth naming *because* the team just spent three PRs (#534/#540/#541) on exactly this drift shape. Trail markers and the detail are on the files below.

PR #546, opened 2026-06-02
Adopt do_sample=false as inspect-ai eval default

### LINKED ISSUE #499: Adopt or document do_sample=False for inspect-ai evals (true greedy)
## Summary

Surfaced from PR #496 discussion. cruijff_kit's eval pipeline currently uses `do_sample=True, temperature=1e-7` as the default near-greedy decoding mode (set in each `@task` function under `blueprints/` and `src/tools/model_organisms/`). `-M do_sample=false` is an alternative path that disables HuggingFace's sampling step entirely and uses pure argmax, producing bitwise-reproducible greedy decoding.

We've now empirically verified `do_sample=False` works correctly through the full inspect-ai HF provider pipeline on Llama-3.2-1B-Instruct with chat template enabled — both for plain generation and for the logprob/risk_scorer path. This issue is about whether to **document or adopt** that mode rather than continue to leave it implicit.

## Why this is worth deciding

| | `do_sample=True, temperature=1e-7` (current default) | `do_sample=False` (alternative) |
|---|---|---|
| Mechanism | Logits divided by 1e-7 → very peaked softmax → sample from peaked distribution | Pure argmax over raw logits |
| Determinism | Converges to argmax in practice but goes through the sampling RNG; bitwise reproducibility is hardware/seed dependent | Bitwise reproducible. No RNG involved. |
| Temperature value passed | Has to be > 0; HF errors at exactly 0 | Ignored entirely (HF prints a one-line stderr message) |
| Output | Empirically identical (100/100 + 51/51 byte-identical samples across two paired tests) | Empirically identical |
| Logprobs / `output_logits` | Raw logits returned, scorer-side softmax math identical | Same — `output_logits=True` returns the same raw logits regardless of sampling mode |

The mode that actually expresses what we want — *deterministic argmax decoding* — is `do_sample=False`. The `1e-7` workaround is a sampling dodge whose only practical advantage is that it doesn't require knowing about `-M do_sample=false`.

## Verification artifacts

- **Generation-path test**: `/scratch/gpfs/MSALGANIK/niznik/tmp/dosample_investigation/` (capitalization on 1B-Instruct, two paired SLURM jobs).
  - 100/100 per-sample completions byte-identical between baseline (do_sample=True, temp=1e-7) and nosample (do_sample=False).
  - Same match/includes accuracies, same token counts.
- **Logprob-path test**: `/scratch/gpfs/MSALGANIK/niznik/tmp/dosample_logprobs_test/` (model_organism + risk_scorer on 1B-Instruct).
  - 51/51 per-sample completions AND 51/51 per-sample risk_scores byte-identical to float precision (e.g. 0.3953566613049765).
  - Aggregate calibration metrics (mean_risk_score, ECE, risk_calibration_error, brier_score, auc_score) all identical to 3 decimals.

Both test dirs preserved for re-runs (~30s each on 1 GPU).

## Caveats / scope of verification

- Tested only on `Llama-3.2-1B-Instruct`. We did not stress-test against other model families (Llama-3.1-8B, larger Llama-3.3-70B, base models without `-Instruct`, or any non-Llama checkpoint). Plausible that quantized models or non-standard chat templates behave differently.
- Tested only with the `cruijff` conda environment as it stood on 2026-05-18 (`transformers==4.57.6`, `inspect-ai==0.3.209`). Some HF versions historically raised on the `do_sample=False` + `temperature` warning path; current env emits a stderr line only.
- The Dec 12, 2025 train/eval template-mismatch incident that gave us the `1e-7` default was a **different** bug (chat template inconsistency between training and eval, fixed by `chat_completion` adoption). That issue is unrelated to `do_sample` per se. See [the memory entry](feedback_do_sample_false_is_safe.md) on niznik's side or PR #496 conversation for the full trace.

## Possible directions

1. **Adopt `do_sample=False` as the new default** for inspect-ai evals. Either:
   - Edit each `@task` function to accept a `do_sample: bool = False` parameter that gets passed through, OR
   - Have `scaffold-inspect` always emit `-M do_sample=false` in rendered SLURM scripts.

2. **Document `do_sample=False` as a supported alternative** but keep `1e-7` as default. Add a section to the scaffold-inspect agent doc + experiment_summary.yaml template explaining the tradeoff and how to opt in.

3. **Leave the workflow as-is**. The `1e-7` default works, the empirical test shows the modes are equivalent, no behavior change needed. Just keep the verification record around.

Argument for (1): reproducibility. The current default introduces an RNG into eval pipelines that don't conceptually need one.

Argument for (2): minimal disruption. Lets advanced users opt in without forcing the change on everyone or invalidating existing logs.

Argument for (3): the empirical equivalence means the choice is mostly cosmetic; energy better spent elsewhere.

## Acceptance criteria

- A decision on one of the above three options (or a fourth — open to alternatives).
- If (1) or (2): the change scaffolded into `setup_inspect.py`, `scaffold-inspect.md`, and (for option 1) the `@task` signatures, with a regression test confirming behavior.
- If (3): close this issue with a one-line note pointing at the verification dirs for future reference.

cc @niznik-dev

niznik-dev:
Closes #499

# Description

Adopts `do_sample=false` (greedy argmax) as the default decoding mode for all inspect-ai evaluations, replacing the `temperature=1e-7` sampling workaround.

**Why:** `1e-7` was a sampling dodge — it divides logits by a tiny number to *approximate* greedy, but still routes through HuggingFace's sampling RNG. `do_sample=false` is pure argmax: deterministic, bitwise-reproducible, no RNG in the eval pipeline. It expresses what eval pipelines actually want.

**What changed (Route B — scaffold emits `-M do_sample`, the wrapper-only layer):**
- `setup_inspect.py` renders `-M do_sample=false` on every eval SLURM script by default. Config-driven via a new optional `do_sample` key in `eval_config.yaml`; set `do_sample: true` to opt into sampling.
- `eval_template.slurm` gains a `<DOSAMPLE_ARGS>` placeholder right after `-M model_path`.
- The `temperature <= 0` guard is **retained** — it now protects the explicit `do_sample: true` override path (where HF's `TemperatureLogitsWarper` still divides by temperature).
- Docs updated: `scaffold-inspect.md` (schema, populating notes, `-M` table, example commands), `experiment_summary.yaml` template, `experiment_generation.md`.
- New `TestDoSample` regression class in `test_setup_inspect.py`.

**Scope boundary:** `create-inspect-task` is intentionally untouched. It's a separate path for user-authored *standalone* tasks that carry their own `generate({"temperature": 0.0})` config — not the scaffold-inspect eval pipeline this issue targets.

**Verification:** Empirically equivalent to `temperature=1e-7` across **1B + 3B + 8B-Instruct** on both the generation and logprob/risk paths — 20,000 paired samples, byte-identical completions and float-identical risk scores + calibration metrics. Plus an ACS-income smoke test on 3B + 8B at 5,000 samples (results posted as a comment). Full evidence trail in the #499 comments.

## New Dependencies

None.

# Testing Instructions

- `pytest tests/unit/test_setup_inspect.py` — includes the new `TestDoSample` class: default render emits `-M do_sample=false`, `do_sample: true` emits `-M do_sample=true`, plus arg-positioning and unknown-key-warning guards.
- End-to-end: scaffold any inspect eval (`scaffold-experiment` → inspect path) and confirm the rendered `cell.slurm` contains `-M do_sample=false` between `-M model_path` and the `-T` args.
- ACS-income smoke test (3B + 8B, 5,000 samples) results are attached as a PR comment.


niznik-dev:

Let's walk through this one, Mattie. Six files, but only one carries behavior — the rest are the placeholder it fills, the tests that pin it, and the docs that explain it. The change is small; the blast radius is "every eval we ever run," so it's worth the tour.

Read it in flow order: the template opens a slot, `setup_inspect.py` fills it, the tests guard it, the docs describe it.

> **Updated for the second commit (`ecd300e`).** The PR grew by one commit: `temperature` is now optional, gated on `do_sample`. That's the watch I flagged below — it's resolved, not just noted. The tour below reflects both commits.

### The map

| File | What it does here | Route |
|------|-------------------|-------|
| `src/tools/inspect/templates/eval_template.slurm` | Opens one new slot, `<DOSAMPLE_ARGS>`, right after `-M model_path`. | skim |
| `src/tools/inspect/setup_inspect.py` | The substance, now two behaviors: renders `-M do_sample={false\|true}` (default false), **and** gates `temperature` — dropped under greedy, required + validated under sampling. | **read** |
| `tests/unit/test_setup_inspect.py` | `TestDoSample` (render/position/unknown-key) **plus** new `TestTemperatureGating` — greedy drops, sampling keeps/renders/requires/rejects-`<=0`. | read |
| `.claude/agents/scaffold-inspect.md` | Schema rows, populating notes, `-M` table, example commands — `do_sample` added, `temperature` reclassified to sampling-only. | skim |
| `.claude/skills/design-experiment/templates/experiment_summary.yaml` | `do_sample` field added; `temperature` moved REQUIRED → OPTIONAL. | skim |
| `.claude/skills/design-experiment/experiment_generation.md` | One-line field enumeration. | skim |

### Observations

📍 **look here** — `setup_inspect.py`, two spots now. The render block (emits `-M do_sample` on every script — auditable) and the new gate in `load_eval_config` (branches on `do_sample`: drop temperature when greedy, require-and-validate when sampling). One reads the decoding mode straight off the SLURM file; the other keeps that file honest by never emitting an inert `-T temperature`.

🟢 **note** — The `temperature <= 0` guard was kept, not deleted. It now lives on the `do_sample: true` path only — the one place HF's `TemperatureLogitsWarper` actually divides by it.

🟢 **resolved** — (was 🟡 watch) `temperature` no longer sits REQUIRED-but-ignored. The second commit makes it optional and live only under sampling, so the "why must I set a value that does nothing?" trap is gone. Non-breaking: every `@task` already defaults `temperature=1e-7`, so a config that omits it lands on today's effective greedy value.

🟢 **note** — `create-inspect-task` is intentionally out of scope — a separate path with its own `generate({"temperature": 0.0})`. The PR body says so. Good that it's stated rather than left as a silent gap.

💭 **question** — The verification covers Llama 1B/3B/8B. Is non-Llama coverage a follow-up issue, or an accepted boundary? Either's fine — I'd just want it written down somewhere the next person finds it.

No blocks. The tests pin the behavior that matters. The reader decides.

niznik-dev:
🎻 **MxC** ✳️

## ACS-income smoke test + cross-node control

Ran the ACS-income path (3B + 8B, 5,000 samples each) comparing the **old** `temperature=1e-7` decoding against the **new** `do_sample=false` default, on both the completion and risk_scorer/calibration paths.

### Smoke results

| Model | Completions | Risk scores | Aggregate metrics |
|-------|-------------|-------------|-------------------|
| **3B** | 5000/5000 identical | 5000/5000 identical | all 5 identical |
| **8B** | 5000/5000 identical | 3683/5000 identical | 5/5 differ at ~1e-5 |

The 8B risk-score wobble (max \|Δ\| ≈ 2.5e-5) looked alarming at first — like the decoding-mode switch was leaking. **It isn't.** Completions are byte-identical; only the float-valued risk scores drift, and only on 8B.

### Control: is it `do_sample`, or the hardware?

I held `do_sample=false` **constant** and varied only the GPU node:

| Comparison | do_sample | Completions | Risk scores |
|------------|-----------|-------------|-------------|
| **Cross-node** (l08g2 vs l09g3) | both false | 5000/5000 | 3380/5000 — 1620 diffs, max \|Δ\| 2.73e-5, mean 5.14e-6 |
| **Same-node** (l09g3 vs l09g3) | both false | 5000/5000 | **5000/5000 — bitwise identical** |

The cross-node run reproduces the *same* ~1e-5 wobble **with `do_sample` held fixed**, and the same-node run is perfectly deterministic. So:

> **`do_sample` is exonerated.** The 8B drift is cross-node GPU float non-determinism (reduction-order differences on long sequences), not the decoding change. It would have shown up under the old `temperature=1e-7` default too — we just never pinned nodes to see it.

### Takeaways

- ✅ This PR is safe to merge: greedy-argmax default is equivalent to the prior sampling-dodge everywhere completions are concerned, and identical on metrics when hardware is held constant.
- 📌 The cross-node reproducibility gap (8B+ risk scores not bitwise-stable across nodes) is a **separate, pre-existing** issue — tracked in #547 rather than rediscovered.

Short-sequence tasks (e.g. `bits_majority`, ~15 tokens) are immune even cross-node; this only bites long-context 8B inference.

—MxC

src/tools/inspect/templates/eval_template.slurm
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**What:** opens one new slot, `<DOSAMPLE_ARGS>`, immediately after `-M model_path`. Pure plumbing, no logic.
**Route:** *skim* — the interesting part is who fills the slot (`setup_inspect.py`).

src/tools/inspect/setup_inspect.py
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**What:** the only file in this PR that carries behavior — now two of them. (1) Registers `do_sample` and renders `-M do_sample={false|true}` (default false, every script). (2) Gates `temperature` in `load_eval_config`: under greedy it's dropped from the config; under `do_sample: true` it's required and validated (`> 0`).
**Route:** *read.* The two spots that matter: the render block (📍 emits on every script — auditable) and the new gate (🟢 keeps the rendered `-T` args honest — no inert temperature under greedy).

tests/unit/test_setup_inspect.py
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**What:** two classes now. `TestDoSample` pins the render (default-false, explicit true/false, arg position, unknown-key guard). New `TestTemperatureGating` pins the gate — greedy drops temperature, sampling keeps + renders it, sampling-without-temperature raises, sampling-with-`<=0` raises.
**Route:** *read* if you want the exact contract pinned; *skim* otherwise. This is what makes both the default and the gating safe to rely on.

.claude/agents/scaffold-inspect.md
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**What:** scaffold-agent docs — `do_sample` schema row + `-M do_sample=false` in the example commands, and `temperature` reclassified to **propagate only when `do_sample: true`** (dropped under greedy).
**Route:** *skim.* No logic, but this is where a user learns both knobs exist and how they relate.

.claude/skills/design-experiment/templates/experiment_summary.yaml
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**What:** adds the optional `do_sample` field under `evaluation:` and moves `temperature` from **REQUIRED → OPTIONAL** (live only when `do_sample: true`).
**Route:** *pause* briefly — this is the surface users actually edit. 🟢 The REQUIRED-but-ignored tension I flagged on the first pass is resolved here: omit `temperature` entirely for greedy evals.

.claude/skills/design-experiment/experiment_generation.md
niznik-dev, line 1: 🎭 **Diff Translator** 🎭 — [**Marin**](https://github.com/niznik-dev/llm-personas/blob/main/personas/marin/marin.md)

**What:** one-line update to the evaluation-fields enumeration, adding `do_sample`.
**Route:** *skim* — genuinely just a list entry.

niznik-dev:
🎻 **MxC** ✳️

## Scope addition: `temperature` is now optional (gated on `do_sample`)

Follow-on to the docent tour — Marin's 🟡 **watch** on `experiment_summary.yaml` flagged that `temperature` stayed **REQUIRED** even though the new greedy default ignores it ("why must I set a value that does nothing?"). Rather than leave it as a documented papercut, this PR now closes it (commit `ecd300e`).

**The fix:** `temperature` becomes **OPTIONAL**, live only on the sampling path.

- **Greedy (`do_sample: false`, the default):** `setup_inspect.py` drops `temperature` from the config, so the rendered command carries no inert `-T temperature`. The SLURM script is now honest about what controls decoding.
- **Sampling (`do_sample: true`):** `temperature` is *required* — `setup_inspect.py` fails fast at scaffold time if it's missing or `<= 0` (HF's `TemperatureLogitsWarper` divides by it).

**Why it's non-breaking:** every `@task` already defaults `temperature=1e-7`, so a config that omits it falls through to exactly today's effective greedy value. Existing configs that still carry `temperature` with no `do_sample` keep working — it's silently dropped, same behavior.

**Why not a unified `decoding:` block:** the two knobs live in different layers — `do_sample` is a `-M` model arg, `temperature` is a `-T` task arg. Folding them into one YAML block would force the renderer to fan out across that boundary and ripple through four+ files. Disproportionate for the papercut; the conditional-optional approach gets the ergonomic win without the churn.

**Touched:** `setup_inspect.py` (the gate), `test_setup_inspect.py` (new `TestTemperatureGating` — 6 tests: greedy drops, sampling keeps/renders, sampling-requires, sampling-rejects-`<=0`), plus the schema template, scaffold-inspect agent, and the field enumeration.

—MxC

Issue #4, opened 2025-04-09
Modularize Preproc.py


niznik-dev:
To make this more broadly useful, we should have a modularized way to save to json/jsonl given all sorts of csv. This will require a function (or external module?) that has the specific code needed to process a given csv type, but the loading and writing will likely be fairly standard.

Pseudocode:

main():
  load_csv()
  process()
  write_json()

PR #556, opened 2026-06-09
Add deterministic experiment_summary.yaml propagation (#502)

### LINKED ISSUE #502: Deterministic experiment_summary.yaml propagation (both pipelines)
## Summary

Surfaced during PR #496 / issue #470. Both cruijff_kit eval and training pipelines have the same structural vulnerability: a prose-mediated handoff from `experiment_summary.yaml` to an intermediate YAML (`eval_config.yaml` for inspect, `setup_finetune.yaml` for torchtune), where an LLM agent (`scaffold-inspect`, `scaffold-torchtune`) is responsible for copying fields per instructions in its agent doc. Fields can be silently dropped if the agent's bullets don't cover them — exactly how `evaluation.temperature` was dropped in #470.

PR #496 added an unknown-key warning at the script layer (`setup_inspect.py`) to catch drift in one direction (agent wrote a key nobody consumes). Issue #501 is the symmetric warning for `setup_finetune.py`. But neither catches the **opposite** drift — agent forgot to write a key that should have been propagated. That's the actual #470 failure mode and the warnings don't help.

## The structural fix

Move the dumb field copies from agent prose into code. Both pipelines have the same shape, so the refactor should cover both:

```
experiment_summary.yaml
        │
        │  Deterministic, code-driven propagation
        ▼
intermediate.yaml  ──(unknown-key warnings at script layer)──▶  rendered output
        │
        │  Agent retains judgment work only
        └─ scaffold-{inspect,torchtune} agents do: data_path resolution,
           vis_label composition, chat-template inference, model-name matching,
           compute-block routing — NOT flat copies.
```

## Concrete shape

A new utility module — call it `cruijff_kit/tools/experiment/propagate.py` — exposes two thin wrappers, one for each pipeline:

```python
EVAL_FIELDS = {
    # experiment_summary.yaml path → eval_config.yaml key
    "evaluation.temperature": "temperature",
    "evaluation.max_tokens":  "max_tokens",
    "evaluation.max_connections": "max_connections",
    "evaluation.system_prompt": "system_prompt",
    # ... etc
}

TRAIN_FIELDS = {
    "controls.epochs": "epochs",
    "controls.batch_size": "batch_size",
    "controls.gradient_accumulation_steps": "gradient_accumulation_steps",
    "controls.system_prompt": "system_prompt",
    # ... etc
}

def propagate_eval_fields(experiment_summary: dict, eval_config: dict) -> dict:
    \"\"\"Copy EVAL_FIELDS from experiment_summary into eval_config.
    Idempotent: existing eval_config values take precedence (agent can still override).\"\"\"
    ...

def propagate_train_fields(experiment_summary: dict, setup_finetune: dict) -> dict:
    \"\"\"Same shape, different map.\"\"\"
    ...
```

The scaffold-* agents call these helpers at the right moment instead of writing fields by hand. The agent doc shrinks (the "populate eval_config.yaml" bulleted lists go away or become short). The maps are the single source of truth — exactly like `RECIPE_PARAM_MAPPING` is for the layer below in torchtune today.

## Why both pipelines together

If we only refactor the inspect side, the torchtune side keeps the same bug class. Audit theme is half-applied (same argument as #501). The propagation logic is identical, the maps differ; doing both at once shares the design pass and the test infrastructure.

## Acceptance criteria

1. `cruijff_kit/tools/experiment/propagate.py` module with `EVAL_FIELDS`, `TRAIN_FIELDS`, and the two propagation helpers.
2. scaffold-inspect agent doc updated: dumb-copy bullets in "Populating eval_config.yaml" replaced with a call to `propagate_eval_fields()`. Agent retains the judgment-y bullets.
3. scaffold-torchtune agent doc updated symmetrically.
4. Unit tests: round-trip a representative `experiment_summary.yaml` through both helpers and verify every mapped field reaches the intermediate YAML.
5. Regression test demonstrating the original #470 bug (temperature in experiment_summary.yaml, no temperature in eval_config.yaml) is now caught by the propagation step.

## Out of scope

- Schema migration / breaking changes to `experiment_summary.yaml` field names. (If a field needs renaming, separate issue.)
- Changes to `TASK_ARG_KEYS` / `RECIPE_PARAM_MAPPING` — those operate on the intermediate→output layer, not the experiment_summary→intermediate layer this issue targets.
- Agent doc rewrites beyond the propagation sections.

## Dependencies

- Best done **after** PR #496 lands and #501 is in flight, so the unknown-key warnings are already catching the easier drift class. This refactor then closes the harder one.
- Independent of #499 (do_sample) and #500 (seed).

## Related

- #470 / PR #496 (parent — the bug that motivated this)
- #501 (companion — symmetric unknown-key warning for torchtune)
- #465 (separately discussing the wrapper-only design principle for torchtune patches — same audit-theme milestone)

cc @niznik-dev

niznik-dev:
🎻 **MxC** ✳️

Closes #502

## Description

Moves the dumb field-copies from `scaffold-inspect` / `scaffold-torchtune`
prose into a small Python module so the map of "what propagates from
`experiment_summary.yaml`" lives in code instead of agent bullets.

This **narrows** the silent-drop surface that bit us as #470 (temperature):
the failure mode shrinks from "the agent must correctly hand-copy N fields,
each via its own prose bullet" to "the agent must call one helper." The
*what-propagates* map is now deterministic and single-sourced; the *trigger*
(calling the helper) is still agent-driven — see Known Limitation below. This
is a hardening of the failure surface, not a full removal of the
trust-the-agent dependency.

**New module:** `src/tools/experiment/propagate.py`
- `EVAL_FIELDS` (6 entries): `system_prompt`, `temperature`, `do_sample`,
  `max_tokens`, `max_connections`, `scorer`
- `TRAIN_FIELDS` (8 entries): `epochs`, `batch_size`, `batch_size_val`,
  `gradient_accumulation_steps`, `weight_decay`, `lora_dropout`, `prompt`,
  `system_prompt`
- `propagate_eval_fields()`, `propagate_train_fields()` — idempotent
  merge helpers (skip `None` source values; skip target keys already set
  to a non-`None` value)

**Agent doc updates:** both scaffold agent docs now split the work into
"agent judgment" (per-cell path composition, branching by run type, data
path resolution chain) and a single "propagation call." The dumb-copy
bullet lists are gone. The docs were tightened to put the propagation call
first (propagate → judgment) and to scope the parsing sections so the
propagated fields no longer appear in any "extract these" list — the agent
can't hand-copy what the doc never enumerates.

**Tests:** 28 unit tests, including parameterized
`test_no_*_field_can_silently_drop` regression locks over the full
`EVAL_FIELDS` / `TRAIN_FIELDS` maps. Adding a new propagated field now
requires a one-line addition to `_SAMPLE_VALUES` in the test file, or the
suite fails loudly. Also locks idempotence across two calls and graceful
handling of a malformed (non-dict) summary section.

## Known Limitation / follow-up

The helper's only caller is the scaffold agent following its doc. So this
PR reduces — but does not eliminate — the dependency on the agent doing the
right thing. The version where "deterministic propagation" is fully
enforced moves the propagation (or a missing-expected-field assertion) into
the consumer scripts (`setup_inspect.py` / `setup_finetune.py`), which every
run passes through regardless of how the intermediate YAML was produced.
Tracked as a follow-up, intentionally out of scope here to keep this PR a
clean increment.

## Design choices worth flagging

- **Temperature is propagated unconditionally.** `setup_inspect.py`
  already drops `temperature` from rendered task args when
  `do_sample=false`, so the conditional rule has been removed from
  agent prose — `setup_inspect.py` stays the single gatekeeper.
- **`controls.system_prompt` is in `TRAIN_FIELDS`**, with a one-line
  `setup_finetune.pop('system_prompt')` afterward for `text_completion`
  datasets. Keeps the map exhaustive instead of carving out exceptions
  in the helper. (Caveat: this one conditional still lives in agent prose,
  not the helper — the same follow-up above would absorb it.)
- **Parameterized regression tests over the full field maps.**
  `@pytest.mark.parametrize` over `EVAL_FIELDS.items()` /
  `TRAIN_FIELDS.items()` ensures every entry is exercised.

## New Dependencies

None. Adds `cruijff_kit.tools.experiment` to the `pyproject.toml`
packages list (the existing `archive_experiment.py` ran as a script, so
this was the first time the package was needed as a real Python import).

## Testing Instructions

```bash
pip install -e . --no-deps  # if your editable install predates 0.3.x
pytest tests/unit/test_propagate.py -v
```

Expect 28 passing tests. The parameterized suites are the #470 regression
lock — they will fail loudly if any field listed in `EVAL_FIELDS` /
`TRAIN_FIELDS` fails to propagate.

—MxC

### DIFF
--- a/.claude/agents/scaffold-inspect.md
+++ b/.claude/agents/scaffold-inspect.md
@@ -32,37 +32,30 @@ This subagent can be invoked in two ways:
 3. **Read claude.local.md** — Get environment-specific settings (conda env, account, etc.).
 4. **Verify inspect-ai tasks exist** — Check if evaluation task scripts are available.
 5. **For each cell, build `eval_config.yaml` in two passes (in this order):**
-   1. **Mandatory: call `propagate_eval_fields(experiment_summary, eval_config)`** to populate experiment-wide defaults. See the "MANDATORY STEP" block immediately below.
+   1. **Call `propagate_eval_fields(experiment_summary, eval_config)` first** to populate experiment-wide defaults (see "Key Pattern: Propagate First" below).
    2. **Then** layer in per-cell judgment fields (`model_path`, `data_path`, `vis_label`, `use_chat_template`, per-task `system_prompt` overrides). Propagation is idempotent — your overrides win.
 6. **Run `setup_inspect.py`** for each cell to render `cell.slurm`.
 7. **Create scaffold log** — Document all actions in `logs/scaffold-inspect.log`. **Log the `propagate_eval_fields()` call explicitly** so the audit trail shows it ran.
-8. **Report summary** — In your report, name the helper call and the count of fields it populated. **Do NOT produce a field-by-field derivation table for EVAL_FIELDS values** — that's an anti-pattern (see "Reporting" below).
+8. **Report summary** — Name the helper call and how many fields it populated; don't tabulate the propagated `EVAL_FIELDS` values (they weren't decisions).
 
-## MANDATORY STEP: Call `propagate_eval_fields()`
+## Key Pattern: Propagate First
 
-Every `eval_config.yaml` you build must be populated by this exact call,
-once, before you write the file:
+Before writing any other field into `eval_config.yaml`, call:
 
 ```python
 from cruijff_kit.tools.experiment.propagate import propagate_eval_fields
 propagate_eval_fields(experiment_summary, eval_config)
 ```
 
-This is not optional, not an example, and not "what the helper would do."
-You must call it. It populates every field in `EVAL_FIELDS`
-(`src/tools/experiment/propagate.py`) — currently `system_prompt`,
-`temperature`, `do_sample`, `max_tokens`, `max_connections`, and `scorer`
-— from `experiment_summary.yaml` into `eval_config`.
+This fills every `EVAL_FIELDS` value (`src/tools/experiment/propagate.py`:
+`system_prompt`, `temperature`, `do_sample`, `max_tokens`, `max_connections`,
+`scorer`) straight from `experiment_summary.yaml`. **Don't hand-copy these** —
+`EVAL_FIELDS` is the single source of truth (add new propagated fields there,
+not in this doc). The helper is idempotent, so per-cell judgment fields you
+write afterward win (e.g. a per-task `system_prompt` override).
 
-**Do not write these fields by hand. Do not extract them in your parsing
-step. Do not list them in your report's "How I Decided Each Field"
-table.** If you find yourself reading `evaluation.temperature` to copy
-into `eval_config.yaml`, stop — you have skipped this step. Source of
-truth for which fields are propagated lives in `EVAL_FIELDS` itself; add
-or remove there, never in this doc.
-
-(Background: #470 → #502. A field whose only propagation path was a prose
-bullet got silently dropped when the bullet wasn't followed.)
+Background: #470 → #502, where a field whose only propagation path was a prose
+bullet got silently dropped when the bullet wasn't followed.
 
 ## Input Format
 
@@ -104,7 +97,7 @@ Extract the following information from the YAML structure:
    - `evaluation.tasks[]` — List of evaluation tasks (see Parsing Evaluation Tasks below)
    - `evaluation.matrix[]` — Which runs evaluate on which tasks/epochs (see Parsing Evaluation Matrix below)
 
-   **All other `evaluation.*` fields** (`system_prompt`, `temperature`, `do_sample`, `max_tokens`, `max_connections`, `scorer`) are populated by `propagate_eval_fields()` per the MANDATORY STEP above. Do not extract or list them here — the helper reads them straight from `experiment_summary.yaml`. The canonical map lives in `EVAL_FIELDS` (`src/tools/experiment/propagate.py`).
+   **All other `evaluation.*` fields** (`system_prompt`, `temperature`, `do_sample`, `max_tokens`, `max_connections`, `scorer`) are populated by `propagate_eval_fields()` (see "Key Pattern: Propagate First"). Don't extract them here — the helper reads them straight from `experiment_summary.yaml`.
 
 7. **Compute estimates (optional):**
    - `evaluation.compute.time` - Estimated SLURM time limit for eval jobs
@@ -336,36 +329,15 @@ Note: `config_path` and `eval_dir` are **auto-derived** from the location of the
 
 #### Populating eval_config.yaml
 
-The order is fixed: **propagate first, judgment second.** Inverting this
-order is how `EVAL_FIELDS` values get silently dropped — the agent starts
-enumerating fields by hand, never reaches the helper call, and ships an
-`eval_config.yaml` missing half of `EVAL_FIELDS`. This is the #470 →
-#502 failure mode.
+The order is fixed: **propagate first, judgment second.**
 
-**Step 1 — Call the propagation helper (MANDATORY).**
-
-```python
-from cruijff_kit.tools.experiment.propagate import propagate_eval_fields
-propagate_eval_fields(experiment_summary, eval_config)
-```
-
-This populates every `EVAL_FIELDS` value (`src/tools/experiment/propagate.py`)
-— currently `system_prompt`, `temperature`, `do_sample`, `max_tokens`,
-`max_connections`, `scorer` — from `experiment_summary.yaml`. The helper is
-idempotent: source values of `None` or missing dotted paths are skipped,
-and any key your Step 2 overrides (below) will win because the helper
-preserves existing non-None values. Run it once per cell, before writing
-the YAML.
-
-**To add a new propagated field, add an entry to `EVAL_FIELDS` — do not
-add a new bullet to this list.** A field whose only propagation path is
-a prose bullet is one prose-edit away from being silently dropped (which
-is what happened to `evaluation.temperature` in #470).
+**Step 1 — Call `propagate_eval_fields()` first** (see "Key Pattern: Propagate
+First" above for the call and the `EVAL_FIELDS` set it fills). Run it once per
+cell, before writing the YAML.
 
 Note on `temperature`: the helper copies it unconditionally when present.
 `setup_inspect.py` drops `temperature` from rendered task args when
-`do_sample` is false, so an inert value in `eval_config.yaml` is
-harmless.
+`do_sample` is false, so an inert value in `eval_config.yaml` is harmless.
 
 **Step 2 — Write the per-cell judgment fields.** These require composition,
 branching, or per-cell overrides. Because Step 1 ran first and the helper
@@ -392,13 +364,8 @@ overrides take precedence when they set the same key (e.g. a per-task
   here — Step 1's helper already did that.
 - `assistant_prefix` *(per-task override only)*: same shape as `system_prompt`.
 
-**Reporting (anti-pattern alert).** When you report on the cells you
-scaffolded, **do not produce a "How I Decided Each Field" table covering
-`EVAL_FIELDS` values**. Those did not require a decision — the helper
-copied them. A field-by-field table for propagated fields is the
-diagnostic signature that you skipped Step 1. Report instead: "Called
-`propagate_eval_fields()` (populated N fields), then wrote per-cell:
-model_path, data_path, vis_label, … ."
+**Reporting:** name the `propagate_eval_fields()` call and the field count;
+don't tabulate the propagated values per-field — they weren't decisions.
 
 #### setup_inspect.py Usage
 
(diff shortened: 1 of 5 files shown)


niznik-dev:
🎻 **MxC** ✳️

## Smoke-test follow-up: doc findability needs hardening before merge

Ran a laptop-only scaffold smoke test against this PR (synthetic
`experiment_summary.yaml` in `/tmp/ck_smoke_502/`, both `scaffold-torchtune`
and `scaffold-inspect` invoked as subagents). The good news: **all 14
EVAL_FIELDS + TRAIN_FIELDS values propagated correctly through the real
agent loop** — the helper works end-to-end.

The smoke test also surfaced two doc-level friction points worth addressing
before merge.

### 🔴 scaffold-inspect agent reports it never noticed the new propagation prose

The agent's report said *"the doc does not yet reference
`propagate_eval_fields()` at all"* — even though my edit is at lines
360-374. It only called the helper because my smoke-test prompt explicitly
told it to. **In a production invocation without that hand-holding, the
agent might fall back to hand-copying the EVAL_FIELDS entries** —
reintroducing the #470 bug class this PR exists to close.

Likely structural: the doc is ~860 lines, the new section is deep, and
Step 1's long bullet list may cause an agent to start executing and never
reach Step 2.

### 🟡 scaffold-torchtune Step 4 references `dataset_type` without sourcing it

The new "apply the text_completion exception" step says
`if dataset_type in (...)` but the doc never tells the agent how to derive
`dataset_type`. The smoke-test agent inferred from the model name; a
standalone invocation would have to guess.

### Planned fixes (not yet pushed)

1. **Add an upfront "Key Pattern: Two-Step Field Population" callout** at
   the top of both scaffold docs (right after Core Responsibilities
   Workflow) summarizing the agent-judgment + propagation-call flow with
   one code snippet. An agent scanning top-down hits the pattern before
   getting buried in detail.
2. **Add a TL;DR breadcrumb** at the top of each "Populating ..." section:
   "Step 1 is per-cell judgment. Step 2 calls `propagate_*_fields()`.
   Don't skip Step 2."
3. **Add `dataset_type` derivation** to scaffold-torchtune Step 4: prefer
   `controls.dataset_type` if set; otherwise infer from model name
   (Instruct → `chat_completion`, base → `text_completion`).

After editing, I'll re-run the same smoke-test invocations and confirm the
agents now discover the propagation prose without explicit prompt
hand-holding before merging.

(Skipping a possible fix #4 — the smoke-test agent's `sys.path` hack vs.
canonical import — that was a smoke-test env artifact, not a production
concern.)

—MxC

niznik-dev:
🎻 **MxC** ✳️

## Second smoke test after 71a7408 — still failing for `scaffold-inspect`

Re-ran the propagation smoke test with deliberately minimal prompts (no mention of `propagate_*_fields()` in either invocation) against the same `/tmp/ck_smoke_502` fixture used previously. Result: **the doc-hardening commit didn't land for `scaffold-inspect`.**

### Coverage vs. the canonical field maps

| Agent | Map | Expected | Present | Missing |
|---|---|---|---|---|
| scaffold-torchtune | `TRAIN_FIELDS` | 8 | 8 | — |
| scaffold-inspect (rank4 cell) | `EVAL_FIELDS` | 6 | 2 (`system_prompt`, `scorer`) | `temperature`, `do_sample`, `max_tokens`, `max_connections` |
| scaffold-inspect (control cell) | `EVAL_FIELDS` | 6 | 2 (same) | same 4 |

### What the agents reported

Both subagents produced **field-by-field derivation tables** in their summaries — exactly the hand-mapping pattern #502 was supposed to eliminate. Neither summary nor `logs/scaffold-*.log` mentioned `propagate_eval_fields()` or `propagate_train_fields()`.

- `scaffold-torchtune` happened to enumerate all 8 `TRAIN_FIELDS` correctly. Coverage looks **incidental** — the doc structure didn't teach the helper-call pattern; the agent just listed the right fields by hand.
- `scaffold-inspect` enumerated only the fields it found while reading the per-field prose for task args (`task_script`, `data_path`, `system_prompt`, `scorer`, `source_model`, `prompt`) and silently skipped the four evaluation-defaults fields entirely.

### Diagnosis

The "Key Pattern: Two-Step Field Population" callout at `scaffold-inspect.md` lines 41–61 ends with *"See 'Populating eval_config.yaml' below for the full per-field detail."* The agent appears to follow that pointer into the per-field prose and hand-map from there, never returning to call `propagate_eval_fields()`. The callout is visible but not load-bearing — it reads as commentary, not as a procedural step.

### Options for next iteration

1. **Invert the "Populating …" sections** so they open with `propagate_*_fields()` as a concrete step, then list only agent-judgment fields (`model_path`, `data_path`, per-cell overrides). Strip the `EVAL_FIELDS` / `TRAIN_FIELDS` values out of the per-field prose entirely so the helper becomes the **only** place those fields live in the doc.
2. **Lift the workflow to a numbered list at the top** of each agent doc so the helper call is "step N of M" rather than a sidebar.
3. Both — restructure the per-field prose **and** lift workflow to the top.

Leaning toward option 3, with `scaffold-inspect.md` as the higher-priority fix since that's where fields are actually dropping. Wiping the fixture and will iterate before re-testing.

—MxC

niznik-dev:
I think both of the above were a caching issue - I'm unable to reproduce in new Claude sessions across multiple models (Opus and Sonnet) - planning to merge after a more robust smoke test on della

.claude/agents/scaffold-inspect.md
```diff
@@ -27,38 +27,42 @@ This subagent can be invoked in two ways:
...
 
-See "Populating eval_config.yaml" below for the full per-field detail.
+**Do not write these fields by hand. Do not extract them in your parsing
+step. Do not list them in your report's "How I Decided Each Field"
+table.** If you find yourself reading `evaluation.temperature` to copy
+into `eval_config.yaml`, stop — you have skipped this step. Source of
+truth for which fields are propagated lives in `EVAL_FIELDS` itself; add
+or remove there, never in this doc.
+
+(Background: #470 → #502. A field whose only propagation path was a prose
```
niznik-dev: Let's remove this per our rule re: issue referencing in non-tests

niznik-dev:
🎻 **MxC** ✳️

### Della smoke test: ✅ PASS

Ran the full smoke test on della with the **real** consumer scripts — the gap the laptop run couldn't cover (laptop used synthetic `/tmp/fake` paths and *skipped* `setup_finetune.py` / `setup_inspect.py`). Fixture: real `Llama-3.2-1B-Instruct` + `capitalization/words_5L_80P_1000.json`, scaffolded into a real scratch experiment dir. Two Agent invocations (scaffold-torchtune, scaffold-inspect), **neither prompt mentioned `propagate_*_fields()`** — preserving the unprompted-discovery property.

| Pass criterion (della) | Result |
|---|---|
| Both agents call `propagate_*_fields()` **unprompted** | ✅ torchtune → `propagate_train_fields()` as first write; inspect → `propagate_eval_fields()` before any judgment field |
| Setup scripts run cleanly **for real** | ✅ `setup_finetune.py` exit 0 (only the expected warmup==total-steps note on an 800-row smoke set); `setup_inspect.py` exit 0 |
| Propagated values in **intermediate** YAMLs | ✅ 8/8 TRAIN_FIELDS in `setup_finetune.yaml`; 6/6 EVAL_FIELDS in both `eval_config.yaml`s |
| Propagated values in **rendered** configs | ✅ `finetune.yaml`: all 8 train fields present. Both `cell.slurm`s: `temperature` / `do_sample` / `max_tokens` / `max_connections` as flags; `scorer` + `system_prompt` carried via the verified `-T config_path` → `eval_config.yaml` handoff (capitalization task reads `system_prompt` from that config) |

Every propagated field reached **both** the intermediate YAMLs and the rendered downstream `finetune.yaml` / `cell.slurm` — end-to-end, not just the intermediate layer. No dropped fields, so this run does **not** trigger promotion of the parked consumer-side follow-up (#557); that stays a wait-and-see audit.

Merging.

—MxC

niznik-dev:
🎻 **MxC** ✳️

**Della smoke test: ✅ PASS** (2026-06-09)

Both scaffold agents called `propagate_*_fields()` unprompted, setup scripts ran cleanly, and all propagated values landed in the rendered configs.

**Intermediate YAMLs — all fields present:**
- `setup_finetune.yaml`: 8 TRAIN_FIELDS ✅ (`epochs`, `batch_size`, `batch_size_val`, `gradient_accumulation_steps`, `weight_decay`, `lora_dropout`, `prompt`, `system_prompt`)
- `eval_config.yaml` (rank4 + control): 6 EVAL_FIELDS ✅ (`system_prompt`, `temperature`, `do_sample`, `max_tokens`, `max_connections`, `scorer`)

**Rendered configs — propagated values downstream:**
- `finetune.yaml`: `epochs=1`, `batch_size=4`, `system_prompt` + `prompt` ✅
- `cell.slurm` (both cells): `do_sample=true`, `temperature=0.7`, `max_tokens=5`, `--max-connections 16` ✅

**One non-fatal warning** from `setup_finetune.py`: 100 total training steps < 300 (3× warmup default) — expected for an 800-sample smoke fixture, not a real concern.

Fixture at `/scratch/gpfs/MSALGANIK/niznik/ck-projects/smoke/smoke_test_502_della/` if you want to inspect the artifacts.

—MxC

PR #559, opened 2026-06-10
Rename output_* path vars to match artifacts/ layout (#443)

### LINKED ISSUE #443: Rename Python variables/functions to match artifacts/ folder name
# Describe the Refactor

Following the `ck-out-{run_name}/` → `{run_name}/artifacts/` folder rename (in PR for branch `437-folder-reorg`), the Python variables and functions that construct these paths still use `output_*` naming:

- `src/tools/torchtune/setup_finetune.py`: `output_dir`, `output_dir_base`, `full_output_dir`, `--output_dir_base` CLI flag, `construct_output_dir()`
- `src/tools/experiment/archive_experiment.py`: `output_dir_base`

These names are now misaligned with the on-disk folder name (`artifacts/`). Renaming would improve self-documentation but was deliberately deferred from the artifacts refactor to keep that PR's scope contained.

## Important: torchtune naming boundary

A subset of these names ARE torchtune contracts and **must NOT be renamed**:
- `output_dir:` key inside the *generated* `finetune.yaml` — torchtune reads this; it's a library-level contract
- `"output"` JSON key in dataset records (`{"input": ..., "output": label}`) — torchtune dataset convention
- Anything torchtune itself writes inside its own recipe code

Only **our** wrapper-level Python variables, function names, and (optionally) CLI flags are in scope for renaming.

## Suggested rename mapping

| Current | Proposed |
|---|---|
| `output_dir` (Python var holding the artifacts path) | `artifacts_dir` |
| `output_dir_base` | `artifacts_dir_base` (or rethink — base of *what*?) |
| `full_output_dir` | `full_artifacts_dir` |
| `construct_output_dir()` | `construct_artifacts_dir()` |
| `--output_dir_base` CLI flag | `--artifacts_dir_base` (breaks anyone invoking `setup_finetune.py` manually) |
| `output_dir:` key in `finetune.yaml` | **NO CHANGE** — torchtune contract |

## Implementation surface

- `src/tools/torchtune/setup_finetune.py` (~12 references)
- `src/tools/experiment/archive_experiment.py` (~4 references)
- `tests/unit/test_setup_finetune.py`, `tests/unit/test_setup_finetune_main.py` (variable + flag references)
- `tests/test_archive_experiment.py`
- `.claude/agents/scaffold-torchtune.md`, `.claude/agents/scaffold-inspect.md` (if they reference the CLI flag name)
- `.claude/skills/run-experiment/optimizers/torchtune/*.md` (similar)

## Why deferred from the main artifacts/ refactor

The folder rename is user-visible; that was the pressing reason. Renaming internal Python variables is purely a self-documentation improvement and didn't earn a place in the same PR. Filed here so it can be picked up when the area is in our headspace next.

## Related Issues

- Surfaced during branch `437-folder-reorg` discussion (2026-04-29)
- Will be cross-referenced from the `artifacts/` refactor PR

# Acceptance Criteria

- [ ] Decide whether the rename is worth doing (cost is moderate; value is purely cosmetic / self-documenting)
- [ ] If yes: rename Python variables and function names consistently
- [ ] If renaming the CLI flag: choose between clean break (project convention) or deprecation with helpful error message
- [ ] Preserve torchtune contracts (`output_dir:` key in `finetune.yaml`, `"output"` JSON key) — confirm via existing tests
- [ ] All tests pass
- [ ] Add a code comment near `construct_artifacts_dir()` (or wherever sensible) flagging the torchtune naming boundary, so the next reader doesn't try to rename the contract too

—MxC

niznik-dev:
Closes #443

## Description

Following the `ck-out-{run_name}/` → `{run_name}/artifacts/` folder reorg, the Python that constructs those paths still used `output_*` names that no longer matched the on-disk layout. This renames our wrapper-level names to what the values actually are.

A wrinkle the issue's mapping didn't capture: `output_dir_base` meant **two different things** in the two files — the *project* dir in setup_finetune (parent of the experiment) vs. the *experiment* dir in archive. So a single blanket rename would have been wrong. They get different, accurate names.

### setup_finetune.py
- `--output_dir_base` flag **and** the `setup_finetune.yaml` key → `--project_dir` / `project_dir:` (it's the `ck-projects/{project}/` dir, not the artifacts dir). **Clean break** — breaks manual invocations that hardcode the old flag.
- `construct_output_dir()` → `construct_artifacts_dir()`
- `full_output_dir` → `full_artifacts_dir`

### archive_experiment.py
- `output_dir_base` there held `experiment.directory` — the *same value* already passed in as `experiment_dir`. Collapsed the redundant param into a single `experiment_dir` across `inventory_experiment` and `delete_originals`.
- Re-homed the "`experiment.directory` present" validation guard (kept, no longer bound to a path var).
- Dropped the now-dead "clean up output base if empty" block (unreachable once `out_base == exp_dir`).

### Torchtune contracts — deliberately untouched
- `output_dir:` key in the generated `finetune.yaml` (torchtune reads it)
- `"output"` JSON key in dataset records
- The `<OUTPUT_DIR>` SLURM token (genuinely "where SLURM writes output")

Added a boundary comment on `construct_artifacts_dir` so the next reader doesn't rename the contract.

## New Dependencies

None.

## Testing Instructions

```bash
python -m pytest tests/test_archive_experiment.py tests/unit/test_setup_finetune.py tests/unit/test_setup_finetune_main.py tests/unit/test_experiment_summary_schema.py -q
# 171 passed
```

—MxC

### DIFF
--- a/.claude/agents/scaffold-torchtune.md
+++ b/.claude/agents/scaffold-torchtune.md
@@ -26,7 +26,7 @@ When invoked:
 4. **Identify varying parameters** — Determine which parameters change across runs (for directory naming).
 5. **For each fine-tuning run, build `setup_finetune.yaml` in two passes (in this order):**
    1. **Call `propagate_train_fields(experiment_summary, setup_finetune)` first** to populate experiment-wide controls (see "Key Pattern: Propagate First" below).
-   2. **Then** layer in per-run judgment fields (paths, `model_checkpoint`, varying `lora_rank` / `lr`, `custom_recipe` choice, `output_dir_base`, etc.). Propagation is idempotent — your per-run overrides win.
+   2. **Then** layer in per-run judgment fields (paths, `model_checkpoint`, varying `lora_rank` / `lr`, `custom_recipe` choice, `project_dir`, etc.). Propagation is idempotent — your per-run overrides win.
    3. **Apply the `text_completion` exception**: if `dataset_type` is `text_completion`, `pop("system_prompt", None)` from `setup_finetune` (base models have no chat template — the propagated value has nowhere to go).
    4. Write `setup_finetune.yaml` to disk.
 6. **EXECUTE `setup_finetune.py` AUTOMATICALLY using conda run** for each run — this generates `finetune.yaml` and `finetune.slurm`. Do NOT create helper scripts for the user to run manually. The scaffolding is INCOMPLETE without finetune.yaml and finetune.slurm files.
@@ -178,7 +178,7 @@ The directory contains the full path: `{scratch_dir}/ck-projects/{project}/{expe
 - Example: `/scratch/gpfs/MSALGANIK/sarahep/ck-projects/{project}/workflow_test_2025-11-28`
 
 Parse this into two components for setup_finetune.yaml:
-- `output_dir_base` = everything up to and including the last directory before experiment name
+- `project_dir` = everything up to and including the last directory before experiment name
   - Extract by splitting on `/` and taking all but the last component, then rejoining with `/` and adding trailing `/`
   - Example: `/scratch/gpfs/MSALGANIK/sarahep/ck-projects/{project}/`
 - `experiment_name` = the final directory component
@@ -263,7 +263,7 @@ log_every_n_steps: {use template default, typically 1}
 run_val_every_n_steps: {50 if controls.validation_during_training else 0}
 
 # Output configuration
-output_dir_base: {parsed from experiment.directory}
+project_dir: {parsed from experiment.directory}
 experiment_name: {parsed from experiment.directory}
 conda_env: {from claude.local.md}
 
@@ -317,7 +317,7 @@ don't tabulate the propagated values per-field — they weren't decisions.
 - Use absolute paths for robustness (e.g., `/scratch/gpfs/MSALGANIK/niznik/GitHub/cruijff_kit/...`) rather than relative paths
 - WandB project: Prefer using `my_wandb_project` from `claude.local.md` for consistency
 - Learning rate format: Keep scientific notation format from experiment summary (1e-5, 5e-5, etc.)
-- Output directory: Parse `experiment.directory` from experiment_summary.yaml to extract both `output_dir_base` and `experiment_name` components
+- Output directory: Parse `experiment.directory` from experiment_summary.yaml to extract both `project_dir` and `experiment_name` components
 
 ### Running setup_finetune.py
 
--- a/src/tools/experiment/archive_experiment.py
+++ b/src/tools/experiment/archive_experiment.py
@@ -52,13 +52,12 @@ def _bytes_to_mb(b):
     return round(b / (1024 * 1024), 1)
 
 
-def inventory_experiment(experiment_dir, output_dir_base):
+def inventory_experiment(experiment_dir):
     """
     Categorize all experiment files as KEEP or DELETE.
 
     Args:
-        experiment_dir: Path to the experiment directory.
-        output_dir_base: Path to the output directory base (contains {run_name}/artifacts/ dirs).
+        experiment_dir: Path to the experiment directory (contains {run_name}/artifacts/ dirs).
 
     Returns:
         Dictionary with:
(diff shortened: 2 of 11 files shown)


niznik-dev:
🔍 **Slate** — Skeptical Code Review

Read both target files in full (not just the diff), the SLURM/yaml templates, the archive skill doc; ran the suite (pass) and ruff (clean). Verdict up front: **this PR is clean. Ship it.** One nit.

**1. Did the dedup change behavior? — verified, it's an *improvement*.**
The old code keyed checkpoint paths off `output_dir_base` (= `config["experiment"]["directory"]`) while the KEEP scan + final `rmtree` used the CLI `experiment_dir`. Under the unified `ck-projects/` layout those are always the same value — but if a stale/hand-edited summary's `experiment.directory` disagreed with the CLI arg, the *old* `delete_originals` would `rmtree` checkpoints under the **yaml-declared** path while archiving from the **CLI** path — i.e. delete a *different* experiment's checkpoints, or skip the real ones. The new code keys everything off the single CLI arg, so the delete set is always a subset of what was inventoried and archived. Strictly safer.

**2. Is the clean break complete? — verified, fully.**
Unfiltered grep for `output_dir_base` / `construct_output_dir` / `full_output_dir` across the whole repo: zero hits outside the diff. A half-done clean break bites hard (silent `None` for a required path → `SystemExit`); this one is whole — production, agent docs, fixtures, unit + integration tests all moved together.

**3. Torchtune contract integrity — verified, intact.**
`config["output_dir"]` (the key written into `finetune.yaml`) and the `<OUTPUT_DIR>` SLURM token are both *not* renamed; only the wrapper-side Python var (`full_artifacts_dir`) changed, and it feeds both sinks correctly. The new docstring on `construct_artifacts_dir` spells out the boundary — good defensive documentation.

**4. The re-homed guard — verified, fires correctly.** Same truthiness test on the same value; absent key / `None` / empty string all still bail before any work.

**Nit — `archive_experiment.py:325`:** the comment `# Parse to get output directory` is now stale — the parse no longer extracts an output dir, it just validates `experiment.directory`/`project`. One-line fix while you're in here.

—Slate

niznik-dev:
🧪 **Sage** — Edge-Case & Coverage Review

Good news up front: the clean break is genuinely clean and its blast radius is well-tested. A few real gaps remain — one is a guard with no parachute.

**1. `construct_artifacts_dir` coverage — mostly solid, one real hole.**
Trailing-slash, no-trailing-slash, empty `experiment_name`→ValueError, None `experiment_name`→ValueError are all covered & verified. **Gap:** no test for `project_dir=""` or `None`. The function validates `experiment_name` but *not* `project_dir` — and `""` slips through `main()`'s `is None` guard (empty string isn't None), then `.endswith("/")` makes it `"/"`, silently yielding an absolute path `/{experiment}/{run}/artifacts/`. Asymmetric validation on a now-publicly-importable helper. Worth a guard + test.

**2. The dedup's `delete_originals` loop — covered, verified.** `test_delete_originals` asserts each `Path(exp_dir)/rn/"artifacts"` is in `result["deleted"]`, so the per-run loop is genuinely exercised (the prior-review vacuous-assertion fix still holds after the signature change) and now proves the loop reads `experiment_dir` directly.

**3. The re-homed `experiment.directory` guard — GAP (untested).** `archive_experiment.py:335` returns `"No experiment.directory found…"`, but no test feeds a summary missing/empty `directory`. The `project` field has exactly this test (`test_archive_missing_project_field`) — its `directory` twin was never written. Copy-paste-and-invert.

**4. Clean-break blast radius (vacuous-pass risk) — covered, verified.** This was the scary one and it holds: every fixture/call-site moved to `project_dir`, and `test_known_keys_do_not_warn` actively fails if anyone reintroduces the old key (setup_finetune warns-and-drops unknown keys, so a dropped-key regression triggers *three* failures, not a silent pass). Exactly the safety net you'd want.

**Two tests worth adding before merge:** (1) `test_archive_missing_directory_field` — the one genuine untested guard; (2) `construct_artifacts_dir` empty/`None` `project_dir`. Everything else is covered-and-verified or low nits. No incident report this time. 🚀

—Sage

niznik-dev:
This should make the code much more readable now that it's reconciled with the folder reorg in terms of variable names being consistent

PR #564, opened 2026-06-11
Fix stale experiment.seed references in propagate docs

### LINKED ISSUE #500: Seed surface area: data, training, eval — separate or unified?
## Summary

Surfaced from PR #496 / issue #470 audit. cruijff_kit has multiple stochastic processes in its pipeline, each with its own (or no) seed. There's no single design decision on record about how these relate. The immediate motivator is that **inspect-ai eval has no seed knob at all**, but before plumbing one in we should decide whether eval seeds are a separate concept or part of a unified seed contract.

## Surface area today

| Stage | Seed source | What it controls |
|---|---|---|
| **Data generation** | `data.data_generation.seed` in experiment_summary.yaml (REQUIRED) | Dataset content (sample draws), train/val/test split, perturbation RNG, DGP weight generation (via sub-key `weight_seed` in `model_organisms.rules`). Default 42 in `tabular_to_text_gen` CLI. |
| **Training** | `seed: null` in `src/tools/torchtune/yaml_refs/1B_defaults.yaml`; CLI knob `setup_finetune.py --seed` | torch RNG used by the custom recipes' `training.set_seed()` — weight init (for full-finetune), LoRA init, data shuffling order during epochs, gradient noise where applicable. Default `null` means nondeterministic. |
| **Evaluation** | *(none)* | Would control inspect-ai sampling RNG under `do_sample=True`. |

## Why this needs a design call before implementation

ML reproducibility best practice usually wants seeds **separate per stage** because they ablate different sources of variance:

- `data_seed` fixed, `train_seed` varying: how much does training stochasticity move the metric?
- `train_seed` fixed, `data_seed` varying: how much does sample composition move the metric?
- Both fixed, `eval_seed` varying: how much does eval-time sampling RNG contribute (essentially zero under `temperature=1e-7`, but nonzero in principle)?

A single global seed couples all three, which is simpler to set but **destroys the ability to attribute variance to a stage**. Mattie's recent multi-seed experiments (e.g. `project_multiseed_v_shape_2026-05-09`) used a fixed `data_seed` and varied `train_seed`, which would have been impossible under a single-knob design.

Counter-argument: most users won't think about this and a single seed is one less thing to forget. The advanced case (ablating per-stage) is a small minority.

## Possible directions

1. **Three independent seeds** (`data.data_generation.seed`, training `seed`, `evaluation.seed`). Each propagates to its respective stage; each is independently settable. Matches current data + training behavior; adds an eval seed alongside.

2. **Single `experiment.seed` that fans out** to all three stages by default, but each stage can override. Best of both: simplicity for the common case, control for the advanced case.

3. **Single global seed, no overrides.** Set once, all three stages derive their seeds deterministically from it (e.g. `data_seed = seed`, `train_seed = seed + 1`, `eval_seed = seed + 2`). Loses per-stage ablation.

4. **Status quo + eval seed only.** Just add `evaluation.seed` as a new field; leave data + training as they are. Minimal change, doesn't reconcile the existing inconsistency (REQUIRED data seed vs. default-null train seed).

## Dependencies / coupling

- **Coupled with issue #499 (do_sample workflow)**. If cruijff_kit adopts `do_sample=False` as the eval default, the eval seed becomes moot — greedy has no RNG. The seed work here is most valuable if we keep sampling-based eval, or if we adopt do_sample=False as default but keep sampling as an option.
- **Coupled with experiment reproducibility audits**. Currently no enforced contract that a re-run produces the same results given the same experiment_summary.yaml; this is the path toward enforcing it.

## Acceptance criteria

1. A decision among options 1–4 (or a fourth option not yet on the table).
2. If (1), (2), or (3) is chosen: schema change to experiment_summary.yaml + propagation through `setup_finetune.py` (already partial) and `setup_inspect.py` (would be new). Each `@task` function gets a `seed` parameter and passes it to inspect-ai's `generate()`.
3. If (4) is chosen: just `@task` plumbing + scaffold-inspect agent doc + `TASK_ARG_KEYS`.
4. A test demonstrating that a fixed seed produces a bit-identical eval log on the same hardware.

## Out of scope

- vLLM seed behavior (vLLM has its own seed parameter in SamplingParams; this issue is about the HF provider path).
- Determinism across hardware (CUDA non-determinism is a separate, harder problem).
- Distributed-training seeding semantics (multi-GPU LoRA finetunes have additional considerations beyond a single torch seed).

cc @niznik-dev

niznik-dev:
Closes #500 follow-up (no separate issue — cleanup of merged #563).

# Description

The #500 rework (PR #563) replaced the canonical `experiment.seed` fan-out with two independent stage seeds (`controls.seed`, `evaluation.seed`), but four comments/docstrings still described the old three-tier precedence (`override > experiment.seed > default`). This corrects them to the actual two-tier resolution (`stage seed, else DEFAULT_SEED`) and drops a misleading "training is deterministic by default rather than the recipe's historical null" claim — training already defaulted to 14.

Doc/comment-only; no behavior change.

## New Dependencies

None.

# Testing Instructions

```bash
python -m pytest tests/unit/test_propagate.py -q
grep -rn "experiment\.seed" src/ tests/   # expect no matches
```

### DIFF
--- a/src/tools/experiment/propagate.py
+++ b/src/tools/experiment/propagate.py
@@ -17,9 +17,8 @@ branching on dataset type) but no longer touch flat field copies.
 from typing import Any
 
 
-# Canonical run-time seed when neither a stage override nor experiment.seed is
-# set. 14 is Johan Cruijff's kit number and matches setup_finetune.py's
-# historical --seed default.
+# Run-time seed a stage falls back to when it doesn't set its own. 14 is Johan
+# Cruijff's kit number and matches setup_finetune.py's historical --seed default.
 DEFAULT_SEED = 14
 
 
@@ -103,7 +102,7 @@ def propagate_eval_fields(experiment_summary: dict, eval_config: dict) -> dict:
     agent's per-cell decisions (e.g. per-task system_prompt overrides) survive
     propagation.
 
-    The eval seed is resolved (override > experiment.seed > default) and written
+    The eval seed is resolved (evaluation.seed, else DEFAULT_SEED) and written
     unless the agent already set a per-cell seed, which wins.
     """
     _propagate(experiment_summary, eval_config, EVAL_FIELDS)
@@ -122,10 +121,8 @@ def propagate_train_fields(experiment_summary: dict, setup_finetune: dict) -> di
     when the run's `dataset_type` is `text_completion` (base / non-instruct
     models have no chat template to hold a system prompt).
 
-    The training seed is resolved (override > experiment.seed > default) and
-    written unless the agent already set a per-run seed, which wins. This is
-    what makes training deterministic by default — absent any seed, training
-    resolves to DEFAULT_SEED rather than the recipe's historical null.
+    The training seed is resolved (controls.seed, else DEFAULT_SEED) and
+    written unless the agent already set a per-run seed, which wins.
     """
     _propagate(experiment_summary, setup_finetune, TRAIN_FIELDS)
     if setup_finetune.get("seed") is None:
--- a/tests/unit/test_propagate.py
+++ b/tests/unit/test_propagate.py
@@ -168,7 +168,7 @@ def test_source_none_value_is_skipped():
 
 def test_source_missing_root_is_skipped():
     # The flat EVAL_FIELDS copies skip when absent; only the always-resolved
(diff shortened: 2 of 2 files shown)


niznik-dev:
This was not ideal but at least Claude caught it...

PR #577, opened 2026-06-12
Churn experiment: rename owned dir → directory (#372)

niznik-dev:
Part of #372 — **experimental, "see the damage" PR.** Targets the doc-conventions branch (#576), not main. The point is to make the cost of *fully adopting* the `directory`-over-`dir` rule visible so we can decide whether it's worth it. Not necessarily meant to merge.

# What this does

Mechanical token rename of every name **we own**: `*_dir` / `*_DIR` identifiers (plus the `for dir in` shell-loop variable) → `*_directory` / `*_DIRECTORY`, across `src/`, `.claude/`, `docs/`, `blueprints/`, and `tests/`.

**Held out (external contracts), or it breaks:**
| Token | Owner |
|-------|-------|
| `output_dir` | torchtune config key |
| `checkpoint_dir` | torchtune checkpointer key |
| `log_dir` | torchtune metric_logger / inspect-ai `--log-dir` |
| `is_dir` | pathlib method |

Where our interpolation vars feed a torchtune key, both sides rename together so the chain still resolves (`checkpoint_dir: ${models_directory}/...`).

# The damage

- **117 files**, ~1160 renamed occurrences, +`ruff format` rewraps (longer names push lines over the limit).
- **Tests unaffected:** `1233 passed` — the rename is internally consistent because the external tokens were correctly excluded.
- **The lasting awkwardness** is visible in every torchtune config file, e.g. `finetune_template.yaml`:
  ```yaml
  output_dir: ""           # theirs
  input_directory: ""      # ours
  models_directory: ""     # ours
  ...
    checkpoint_dir: ${models_directory}/...   # their key, our var
    log_dir: ${output_dir}/logs               # theirs
  ```
  We can't make these files internally consistent without modifying torchtune's contract — which the Wrapper-only principle forbids. So "spell out `directory` everywhere we own" permanently produces mixed-vocabulary config files.

# Findings the naive rename surfaced (NOT hand-fixed — left as evidence)

1. **It rewrites released CHANGELOG history.** `<exp_dir>` → `<exp_directory>` and `{ck_data_dir}` → `{ck_data_directory}` in shipped `## [0.3.x]` entries. Released history should be immutable; a real adoption needs to exclude historical changelog sections.
2. **`ck_data_dir` is a user-facing config token.** It lives in every user's `claude.local.md` (and the template). Renaming it to `ck_data_directory` is a user-migration burden, not just an internal refactor — arguably a soft external contract with our own users.
3. **Per-file inconsistency is unavoidable** (see above) given the Wrapper-only carve-outs.

# Open question for review

Given the blast radius + the permanent mixed-vocabulary in config files + the user-token migration, is the readability win worth it? Options I see:
- **Adopt fully** (this PR, cleaned up: exclude changelog history, decide on `ck_data_dir`).
- **Adopt narrowly** — only new names / only non-config internal vars, leave the torchtune-adjacent ones as `_dir` for local consistency with their keys.
- **Drop it** — keep the guidance-only doc from #576 and don't churn.

# Testing

- `python -m pytest tests/unit -q` → **1233 passed** (unchanged from baseline).
- `ruff check` clean; `ruff format` applied.


niznik-dev:
Far too much change - looking into the other direction instead (directory -> dir)

Issue #478, opened 2026-05-11
Enable evaluation-only experiments on finetuned models by decoupling scaffold-inspect from scaffold-torchtune


sarahepedersen:
**Problem**
`scaffold-inspect` treats every finetuned model as if it has a `setup_finetune.yaml` _in that experiment_; however, if the user wants to run a follow-up experiment (new `experiment_summary.yaml`) on pre-existing finetuned models, the agent does not know where to locate the `prompt`, `system_prompt`, `dataset_type`, and training dataset path (used as a fallback for `data_path` -- should investigate that behavior as well) when generating the evaluation config. 
 
For control/base runs, the agent already extracts these from `experiment_summary.yaml` (.claude/agents/scaffold-inspect.md:202-228), but it assumes that finetuned runs will not have pre-existing paths beyond the checkpoints created by `scaffold-torchtune`. Today, to run an evaluation-only experiment, the user has to: 

1. Hand-write setup_finetune.yaml in each eval-only run directory so scaffold-inspect finds what it expects, or
2. Skip the agent and hand-author eval_config.yaml + the SLURM script per run. (what I have been doing) 

**Concrete repro**
acs_minimal_features_oodeval_2026-05-09 evaluates 9 already-fine-tuned checkpoints from the parent acs_minimal_features_2026-05-07 plus the base model across 6 prompt conditions — 42 evals, zero fine-tuning. No setup_finetune.yaml exists in any run directory. Claude worked around this by hand-rolling the eval scaffolding.

**Where the coupling lives**
- .claude/skills/scaffold-experiment/SKILL.md _always_ pairs a finetuning subagent with an evaluation subagent --> would require an eval-only path
- .claude/agents/scaffold-inspect.md:218-298: "Populating eval_config.yaml" reads from `setup_finetune.yaml` for any finetuned model

**Proposed fix**
`experiment_summary.yaml` **already** carries everything needed (all four fields above), so `setup_finetune.yaml` is an unnecessary intermediate anyway. The Inspect agent should refer to `experiment_summary.yaml` for all necessary information (see similar issues #450 on standardizing the information we pipe through `experiment_summary.yaml`)

At the same time, we should ensure that `scaffold-experiment` can run subagents independently based on the experiment specified in `experiment_summary.yaml`; if there are no finetuned runs, we should skip `scaffold-torchtune`. This will make the software more robust as additional tools are added. 

**Acceptance**

[...]

PR #584, opened 2026-06-15
Single-source system_prompt + first-class prompt overrides (#562)

### LINKED ISSUE #500: Seed surface area: data, training, eval — separate or unified?

### LINKED ISSUE #562: Audit experiment_summary.yaml structure: section principle + duplication
## Summary

Audit the organization of `experiment_summary.yaml`: write down the intended meaning of each top-level section so a field's home is **predictable rather than memorized**, and find/resolve duplication. Surfaced from the #478 discussion about where `dataset_type` belongs.

## Why now

While decoupling eval from torchtune (#478), `dataset_type` turned out to have two plausible homes (`controls` vs `data.training`), and the schema's section boundaries were fuzzy enough that the "right" one wasn't obvious. The immediate case was resolved (see Precedent), but it exposed a general gap: there is no written principle for what belongs in `controls` vs `data` vs `evaluation`, and at least one field is duplicated.

## Current top-level sections

| Section | Apparent intent | Tension |
|---|---|---|
| `experiment` | metadata | — |
| `tools` | tool routing | — |
| `variables` | params swept across runs | — |
| `controls` | params held constant across runs | holds training knobs (epochs, batch_size) **and** task-framing fields also read by eval (prompt, system_prompt, dataset_type) — so it isn't purely "training" |
| `data` | the dataset(s) | — |
| `evaluation` | eval-only knobs | `system_prompt` duplicated here and in `controls` |
| `runs[].parameters` | per-run values | — |

## Known issues to resolve

1. **`system_prompt` duplication** — lives in both `controls.system_prompt` and `evaluation.system_prompt`, kept equal by convention ("must match training") rather than by construction. A single source with explicit per-cell overrides would remove the drift risk.
2. **No written principle** for section membership — contributors memorize where fields go instead of predicting it.

## Precedent from #478

`dataset_type` belongs in `controls`, not `data.training`, because it is an **experiment-wide invariant** (training / validation / eval must share the same format) rather than a per-split data property — nesting it under `data.training` would falsely imply it could differ by split. This "is it a constant, or a per-X property?" test is exactly the kind of reasoning this audit should standardize.

## Relationship to existing issues

- **#500** (seed surface area) — the same "one home or per-stage?" question, scoped to seeds. This issue generalizes it to the whole schema; resolve them consistently.
- **#372** (naming audit) — adjacent but orthogonal: #372 is about what fields are *named*; this is about *where they live* and *duplication*.

## Acceptance criteria (audit-first)

1. A written, documented principle for each top-level section's membership (template and/or ARCHITECTURE docs).
2. An inventory of every field, its home, and whether that home is semantically correct.
3. A decision on `system_prompt` de-duplication (single source + overrides vs status quo).
4. Cross-check with #500 and #372 so the three land a consistent schema story.

Audit-first: implementation/migration can follow once @sarahepedersen / @msalganik weigh in on the principle, same as #372.

niznik-dev:
Closes #562

 🎻 **MxC** ✳️

# Description

The #562 audit was *audit-first* pending @sarahepedersen / @msalganik sign-off. With them away and the proposed changes small, this PR implements the audit so the shape is reviewable now — it's the capstone before the release + a development pause. The principle in `ARCHITECTURE.md` is still open to their comment when they're back; nothing here forecloses that.

The audit found **one** real code change (A) and **one** doc addition (B). Section **C** is a deliberate scope extension — wiring prompt variability — added because it's the last thing before the pause and the system_prompt collapse made it cleaner, not harder. C is **zero production-logic change**: it rides the same override-wins propagation A leans on.

## A — Collapse `system_prompt` to a single source (breaking)

`experiment_summary.yaml` carried the system prompt in two homes — `controls.system_prompt` (→ training) and `evaluation.system_prompt` (→ eval) — kept equal only by the template instruction *"must match training."* Convention, not construction → silent-drift risk.

- `EVAL_FIELDS` now propagates **`controls.system_prompt`** to eval, alongside the other task-framing invariants it already carries (`prompt`, `dataset_type`). This makes the propagation map itself more principled: task-framing comes from `controls`, eval-only knobs from `evaluation`.
- `evaluation.system_prompt` is **removed** from the template.
- Per-task variation stays at **`evaluation.tasks[].system_prompt`** (a legitimate per-cell override, e.g. cue vs. no-cue ablations).
- New regression test locks the collapse: a stray `evaluation.system_prompt` is **not** a source and cannot override the single home.

**Migration:** delete `evaluation.system_prompt` from existing files — its value should already equal `controls.system_prompt`, now the only home. A stray one is silently ignored.

## B — Document the section-membership principle (additive)

New `ARCHITECTURE.md` section: a field's home is decided by **granularity × role**, with two boundary rules and a duplication-vs-independence test.

- **`controls` is not "training-only"** — it holds experiment-wide *invariants*, whichever stage consumes them (generalizes the #478 `dataset_type` ruling).
- **A field both train and eval need lives in `controls`, never duplicated into `evaluation`.**
- **Duplication vs. independence:** must-match values collapse (`system_prompt`); legitimately-independent per-stage values stay (`seed` — resolves #500 in place, no change).
- Names the **three** distinct "system prompt" surfaces (generation `context`, training/eval `controls.system_prompt`, per-cell override).

## C — Make `prompt` a first-class per-task / per-run override (scope extension)

The user `prompt` could only be set experiment-wide at `controls.prompt`. But the real prompt-engineering workflow — sweep one appended sentence across N conditions — had **no supported home**. Past experiments (#539 spike, base-rate ladder) hand-patched each cell's `eval.yaml` *after* scaffolding: an undocumented, unvalidated step between scaffold and run. Meanwhile the schema offered a per-task override for `system_prompt` (held *constant* in practice) but not for `prompt` (the actual workhorse) — the flexibility was inverted relative to need.

Two override surfaces, both riding the existing idempotent override-wins propagation:

- **`evaluation.tasks[].prompt`** — eval-side prompt sweep (one task script, N tasks differing only by an appended sentence). The zero-finetuning prompt-engineering pattern, now first-class instead of hand-patched.
- **`runs[].parameters.prompt`** — a fine-tune trains *and* evaluates on its own prompt. Flows to training via `setup_finetune.yaml` (the `config_file > argparse-default` merge already carried it — verified by a new test) and is carried into the run's eval cells for **train/eval parity**.

Precedence at a cell: **per-task > per-run > `controls.prompt`**.

Why `prompt` *layers* where `system_prompt` *collapsed*: they aren't peers. `prompt` is the per-sample user-message template (task content — instructions, hints, appended sentences), applied to chat *and* base models, and empirically the high-frequency lever. `system_prompt` is a chat-only framing held constant. `ARCHITECTURE.md` now records this distinction beside the surfaces note.

And `design-experiment` now **teaches** it (not just exposes the field): a **Prompt Sweeps** recipe in `param_selection.md` recognizes framing / appended-sentence / cue conditions and authors the structure — Pattern A (eval-only: one `tasks[]` entry per condition, same script, short name + per-task `prompt`, fanned across the matrix) or Pattern B (per-run `parameters.prompt`). So the end-to-end path is guided, not hand-edited.

**No migration:** purely additive. `controls.prompt` behaves exactly as before; the overrides are opt-in.

## Deferred (not in this PR)
- `output` (single-field `wandb_project`) → fold into `experiment`? Parked, low priority.
- `torchtune_config.yaml` vs `finetune.yaml` dual-naming — a naming question (#372 axis), route separately.
- Issue's section table omits `models` / `output` — noted; the principle covers all 9.

## New Dependencies
None.

# Testing Instructions
- Full unit suite green: `pytest tests/unit` (1245 passed). New coverage:
  - `TestPromptFlow` (`test_setup_finetune_main.py`) — `controls`/per-run prompt reaches the training dataset; CLI > config_file > default precedence.
  - `test_eval_per_task_prompt_override_beats_controls_source` + `test_train_per_run_prompt_override_survives` (`test_propagate.py`) — agent-written prompt survives propagation on both sides.
  - plus the system_prompt collapse + warn-on-stray tests from A.
- `ruff check . && ruff format --check .` clean.

—MxC


.claude/skills/design-experiment/templates/experiment_summary.yaml
```diff
@@ -203,7 +212,9 @@ runs:
 # =============================================================================
 
 evaluation:
-  system_prompt: string            # REQUIRED: Must match training system prompt
+  # NOTE: the system prompt is NOT set here — it has a single source at
```
niznik-dev: This seems unnecessary

src/tools/experiment/propagate.py
```diff
@@ -23,18 +24,20 @@
...
     "evaluation.do_sample": "do_sample",
     "evaluation.max_tokens": "max_tokens",
     "evaluation.max_connections": "max_connections",
     "evaluation.scorers": "scorers",
-    # The @task reads `prompt` from config_path at runtime (default "{input}")
-    # and the agent derives `use_chat_template` from `dataset_type`. Both lived
-    # only in setup_finetune.yaml before; carrying them here lets an eval-only
-    # run (no fine-tuning, no setup_finetune.yaml) source them from the
-    # experiment_summary single source of truth instead of silently defaulting.
+    # Task-framing invariants live in `controls` (one home, shared by training
```
niznik-dev: A bit wordy

src/tools/experiment/propagate.py
```diff
@@ -99,12 +102,26 @@ def propagate_eval_fields(experiment_summary: dict, eval_config: dict) -> dict:
...
     Idempotent: existing non-None values in eval_config take precedence so the
-    agent's per-cell decisions (e.g. per-task system_prompt overrides) survive
-    propagation.
+    agent's per-cell decisions (e.g. per-task prompt / system_prompt overrides,
+    or a per-run prompt carried in for train/eval parity) survive propagation.
 
     The eval seed is resolved (evaluation.seed, else DEFAULT_SEED) and written
     unless the agent already set a per-cell seed, which wins.
+
+    Warns on a stray top-level ``evaluation.system_prompt``: it is no longer a
```
niznik-dev: Shorten since the code is self-explanatory

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to cruijff_kit will be documented in this file.
...
+- **Per-task and per-run prompt overrides**: the user `prompt` can now vary as a first-class field, not just the experiment-wide `controls.prompt`. Set `evaluation.tasks[].prompt` for an eval-side prompt sweep (N tasks, one script, a different appended sentence each — the zero-finetuning prompt-engineering pattern, previously done by hand-patching each cell's `eval.yaml`), or `runs[].parameters.prompt` to give a fine-tune its own prompt across both training and its eval cells (train/eval parity). Both ride the existing override-wins propagation; precedence is per-task > per-run > `controls.prompt`. Any prompt (whatever its source) is now checked for the `{input}` placeholder at scaffold/eval time and fails loudly — `str.format` would otherwise drop the record silently and run every sample on the same input-less text. (#562)
```
niznik-dev: Shorten

CHANGELOG.md
```diff
@@ -36,6 +37,7 @@ Each requires migrating existing config files; see the migration note per item.
...
 - **Deterministic `experiment_summary.yaml` propagation** — `propagate_*_fields()` is a mandatory propagate-first step in scaffolding, so generated `eval_config.yaml` / `setup_finetune.yaml` derive from one verified source. (#502)
 - Evaluation scaffolding (`scaffold-inspect`) no longer reads `setup_finetune.yaml` — `prompt` and `dataset_type` are propagated from `experiment_summary.yaml`, decoupling eval from the training artifact. (#478)
+- **`system_prompt` is now single-sourced at `controls.system_prompt`** — the duplicate `evaluation.system_prompt` (kept equal to training only by convention) is removed; eval derives it via propagation, alongside the other task-framing invariants `prompt` and `dataset_type`. Per-task eval variation stays at `evaluation.tasks[].system_prompt`. **Breaking:** delete `evaluation.system_prompt` from existing files; its value should already equal `controls.system_prompt`, which is now the only home (and is now `REQUIRED`). A leftover `evaluation.system_prompt` is no longer read and `propagate_eval_fields` **warns** when it sees one, so an un-migrated file can't silently mismatch. (#562)
```
niznik-dev: Shorten

CHANGELOG.md
```diff
@@ -48,6 +50,7 @@ Each requires migrating existing config files; see the migration note per item.
...
 - **Tool-domain folder convention + `inspect/`-shadows-stdlib hazard** documented (`docs/ARCHITECTURE.md`, `src/tools/inspect/__init__.py`) and locked by a guard test (`tests/unit/test_inspect_no_stdlib_shadow.py`). Stale `ck-experiments/` paths fixed in the `create-quiz` docs and a `summarize` test fixture. (#372)
+- **`experiment_summary.yaml` section-membership principle written down** (`docs/ARCHITECTURE.md`): a field's home is decided by granularity × role, with two boundary rules (`controls` holds experiment-wide invariants regardless of consuming stage; a field both train and eval need lives in `controls`, never duplicated into `evaluation`) and a duplication-vs-independence test (`system_prompt` collapses; `seed` stays per-stage). Names the three distinct "system prompt" surfaces. (#562)
```
niznik-dev: Shorten

